PR #6, opened 2020-06-11
Enhance search for GCMD vocabulary values

### LINKED ISSUE #4: Sentinel 2 platforms and instruments are not created from GCMD
The Sentinel 2 datasets from Copernicus scihub are not correctly associated with the corresponding GCMD platforms and instruments.

aperrin66:
Closes #4 
By using pythesint `search_gcmd_*_list()` methods, it is easier to find the correct GCMD vocabularies values from the metadata.

### DIFF
--- a/metanorm/utils.py
+++ b/metanorm/utils.py
@@ -10,18 +10,17 @@ import pythesint as pti
 UNKNOWN = 'Unknown'
 
 
-def get_gcmd_provider(potential_provider_attributes):
+def get_gcmd_provider(potential_provider_attributes, additional_keywords=None):
     """
-    Gets a GCMD provider from a name and/or URL, otherwise generate a GCMD provider-like data
-    structure
+    Get a GCMD provider from a name and/or URL, otherwise return None
     """
     provider = None
     for attribute in potential_provider_attributes:
-        try:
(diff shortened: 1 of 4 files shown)


metanorm/utils.py
```diff
@@ -46,50 +42,105 @@ def get_gcmd_like_provider(name=None, url=None):
...
 
+def gcmd_search(vocabulary_name, keyword, additional_keywords=None):
+    """
+    Search for GCMD objects using the provided vocabulary name and keywords.
+    Returns None if nothing was found.
+    """
+    pti_search_method = getattr(pti, f"search_gcmd_{vocabulary_name}_list")
+    pti_get_method = getattr(pti, f"get_gcmd_{vocabulary_name}")
+
+    gcmd_instrument = None
```
akorosov, line 92: It is not specific to instrument only, right? Why is it used here for instrument only? Or is it just a variable names `*_instrument` and it will work with everything?  Please change the name then.

aperrin66: My bad, I wrote it with instruments in mind and then forgot to change the variable names.

aperrin66: It's fixed

Issue #7, opened 2020-07-02
Add new normalizer based on filename


akorosov:
Develop Sentinel1FilenameMetadataNormalizer that can retrieve

* entry_id
* platform
* instrument
* time_coverage_start
* time_coverage_end
* provider

From filename

Example filename:
ftp://ftp.nersc.no/nansat/test_data/sentinel1_l1/S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1.zip

And (incomplete) example of unittest
```python
class SentinelFilenameMetadataNormalizerTests(unittest.TestCase):
    """Tests for the SentinelFilenameMetadataNormalizer """

    def setUp(self):
        self.normalizer = normalizers.SentinelFilenameMetadataNormalizer([])

    def test_platform(self):
        """ shall return platform from SentinelSAFEMetadataNormalizer """
        attributes = {'filename': 'S1A_...'}
        self.assertEqual(self.normalizer.get_platform(attributes), pythesing_like_representation('Sentinel-1A'))

```

akorosov:
Naming conventions:
https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions

akorosov:
For sentinel-1 the instrument is always instrument = 'C-SAR'

akorosov:
provider is always ESA

PR #8, opened 2020-07-06
#7 adding new normalizer for filename and some tests of it

aanersc:


aanersc:
@akorosov @aperrin66 
Please check the normalizer as well as some tests for it.

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class SentinelFilenameInterpreterNormalizer(BaseMetadataNormalizer):
+    """ Normalizer for extraction of information from the filename """
+
+    def format_checker(self, raw_attributes):
+        """ For more information, please check the format of filename to be correct as declared in
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        for i in [3, 6, 11, 16, 32, 48]:
+            assert raw_attributes['entry_id'][i] == '_'
```
akorosov, line 23: 1. It is more efficient and easier to extend to use regex:
```suggestion
        import re
        re.match('S1\D_\D{2}_\D{4}_\d\D{3}_\d{8}T\d{6}_\d{8}T\d{6}', filename)
```
Read more about regex here:
https://docs.python.org/2/howto/regex.html#regex-howto
and here:
https://docs.python.org/2/library/re.html

2. What happens if format checking fails? It will raise an error. But is the base class ready for that? I guess not. At leas in other normalizers no error is raised. Instead a None values is returned. BTW, in the tests it is not checked what happens when the normalizer tries to process incorrect data.

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def string_cutter(self, raw_attributes, part):
```
akorosov, line 25: It is a good decision to have a single function to perform a very similar operation.
I would probbaly call it `cut_string`, though. Try to call functions with verb coming first.

metanorm/normalizers/filename_interpreter.py
akorosov: re.match will return None if the string doesn't match and a regexp match object otherwise.

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+        """ Cuts the filename string based on incoming part """
+        if part == 'platform':
+            return raw_attributes['entry_id'][0:3]
+        elif (part == 'instrument' or part == 'provider'):
+            return raw_attributes['entry_id'][0:2]
+        elif part == 'time_coverage_start':
+            return raw_attributes['entry_id'][17:32]
+        elif part == 'time_coverage_end':
+            return raw_attributes['entry_id'][33:48]
```
akorosov, line 34: 1. It is a quite rigid approach that is hard to adapt for other sentinel-1 filenames. E.g. for level 2, when size of filename parts changes. Again regex can help a lot here. For example for time coverage:
`re.findall('_\d{8}T\d{6}_\d{8}T\d{6}_', raw_attributes['entry_id'])` will always return
`['_20150702T172954_20150702T173054_']`
and for platform/instrument `re.findall('S1\D_', ' raw_attributes['entry_id'])`
will return either `['S1A_'] or ['S1B_']`

2. using regex opens possibility to easily DRY. Define a dictionary that defines regex and slices:
```python
cutter = dict(
    instrument = dict(regex = 'S1\D_', s0 = 0, s1 = 3),
    time_coverage_start = dict(regex = '_\d{8}T\d{6}_\d{8}T\d{6}_', s0 = 1, s1 = 14),
}
```
and then use it once:

```python
return  re.findall(cutter[part]['regex'], raw_attributes['entry_id'])[cutter[part]['s0']:cutter[part]['s1`]]]
```

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable platform based on the filename """
+        if set(['entry_id']).issubset(raw_attributes.keys()):
+            self.format_checker(raw_attributes)
+            platform_str = self.string_cutter(raw_attributes, 'platform')
+            if platform_str.upper() == 'S1A':
+                return utils.get_gcmd_platform('SENTINEL-1A')  # 'SENTINEL-1A'
+            elif platform_str.upper() == 'S1B':
+                return utils.get_gcmd_platform('SENTINEL-1B')  # 'SENTINEL-1B'
```
akorosov, line 52: Can be replaced with dictionary 
`platform_map = dict(S1A = 'SENTINEL-1A', S1B = 'SENTINEL-1B')`

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            return None
+
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the filename """
+        if set(['entry_id']).issubset(raw_attributes.keys()):
+            self.format_checker(raw_attributes)
+            instrument_str = self.string_cutter(raw_attributes, 'instrument')
```
akorosov, line 60: These lines are repeated in every method. Nice place to optimize!
Can you add a method that calls issubset, format_checker and string_cutter? then you have to call only this function in each method, and then comes unique part for each method.

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+
+import metanorm.utils as utils
+
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class SentinelFilenameInterpreterNormalizer(BaseMetadataNormalizer):
```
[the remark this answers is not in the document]
akorosov: Then I would also rename the module to sentinel1_identifier.py

metanorm/normalizers/filename_interpreter.py
aanersc: string_cutter is a good name. Kindly, I think there is no need to change it.

aperrin66: I agree with Anton: *string_cutter* would be a name for a class, not a method

metanorm/normalizers/filename_interpreter.py
aperrin66: You can also use regex groups to directly generate a dictionary, here is an example:

```python
import re
f = 'S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1'
m = re.match(r'^(?P<platform>S\d[AB])_(?P<mode>[A-Z]{2})_(?P<type>[A-Z]{3})(?P<resolution>[A-Z])_(?P<processing_level>[12])(?P<class>[SA])(?P<polarisationtion>[A-Z]{2})_(?P<start>\d{8}T\d{6})_(?P<end>\d{8}T\d{6})_(?P<orbit>\d{6})_(?P<mission_id>[A-Z0-9]{6})_(?P<product_id>[A-Z0-9]{4})', f)
print(m.groupdict())
```
This would print:
```
{'platform': 'S1A', 'mode': 'EW', 'type': 'GRD', 'resolution': 'M', 'processing_level': '1', 'class': 'S', 'polarisationtion': 'DH', 'start': '20150702T172954', 'end': '20150702T173054', 'orbit': '006635', 'mission_id': '008DA5', 'product_id': '55D1'}
```

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        import re
+        re.match('S1\D_\D{2}_\D{4}_\d\D{3}_\d{8}T\d{6}_\d{8}T\d{6}', filename)
+
+    def string_cutter(self, raw_attributes, part):
+        """ Cuts the filename string based on incoming part """
+        if part == 'platform':
+            return raw_attributes['entry_id'][0:3]
+        elif (part == 'instrument' or part == 'provider'):
+            return raw_attributes['entry_id'][0:2]
```
aperrin66, line 30: The first two characters represent neither the instrument nor the provider.
Unless I am mistaken the instrument is always SAR for Sentinel-1.
The provider (ESA) would be characterized by the fact that the whole identifier matches the expected pattern (which is one more reason to use regular expressions).

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
+"""Normalizer for the interpretation of file name convention"""
+
+import logging
+
+#import dateutil
```
aperrin66, line 5: please don't leave commented imports here

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+import dateutil.parser as ps
```
aperrin66, line 6: Why use an alias? `dateutil.parser` is not that long to type and it's much clearer in the code

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            return None
+
+    def get_provider(self, raw_attributes):
+        """ returns the suitable provider based on the filename """
+        if set(['entry_id']).issubset(raw_attributes.keys()):
+            self.format_checker(raw_attributes)
+            provider_str = self.string_cutter(raw_attributes, 'provider')
+            if provider_str.upper() == 'S1':
+                return utils.get_gcmd_like_provider('ESA')  # 'ESA'
```
aperrin66, line 92:
```suggestion
                return utils.get_gcmd_provider(['ESA/EO'])
```

tests/normalizers/test_filename_interpreter.py
```diff
@@ -0,0 +1,50 @@
...
+            {'entry_id': 'S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1'})
+        self.assertDictEqual(
+            result_normalization['platform'],
+            OrderedDict([
+                ('Category', 'Earth Observation Satellites'),
+                ('Series_Entity', 'SENTINEL-1'),
+                ('Short_Name', 'SENTINEL-1A'),
+                ('Long_Name', 'SENTINEL-1A')
+            ])
+        )
```
aperrin66, line 50: Please make sure to test all the methods of the normalizer

metanorm/normalizers/filename_interpreter.py
akorosov: Wow! That's beautiful! I forgot this functionality. To be honest regex is not my strongest side. I'll use this code as a reference ;)

aanersc:
@akorosov 
@aperrin66 
4f984c9 is kindly ready to merge if you don't have any other comments.

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,103 @@
...
+    """ Normalizer for extraction of information from the filename """
+
+    def check_format(self, raw_attributes_entry_id):
+        """ For more information, please check the format of filename to be correct as declared in
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        return re.match('S1\D_\D{2}_\D{4}_\d\D{3}_\d{8}T\d{6}_\d{8}T\d{6}', raw_attributes_entry_id)
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        m = re.match(r'^(?P<platform>S\d[AB])_(?P<mode>[A-Z]{2})_(?P<type>[A-Z]{3})(?P<resolution>[A-Z])_(?P<processing_level>[12])(?P<class>[SA])(?P<polarisationtion>[A-Z]{2})_(?P<time_coverage_start>\d{8}T\d{6})_(?P<time_coverage_end>\d{8}T\d{6})_(?P<orbit>\d{6})_(?P<mission_id>[A-Z0-9]{6})_(?P<product_id>[A-Z0-9]{4})', raw_attributes_entry_id)
```
aperrin66, line 26: typo: "polarisation**tion**"
Be careful when copy-pasting ;)

aanersc: This is nontouch part of the code! Good notice!

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,103 @@
...
+        if set(['entry_id']).issubset(raw_attributes.keys()):
+            match_result = self.check_format(raw_attributes['entry_id'])
+            if match_result is not None:
+                platform_str = self.cut_string(raw_attributes['entry_id'], 'platform').upper()
+                platform_map = dict(S1A = 'SENTINEL-1A', S1B = 'SENTINEL-1B')
+                return utils.get_gcmd_platform(platform_map[platform_str])
+            else:
+                return None
+        else:
+            return None
```
akorosov, line 51:
```suggestion
    def create_attr_string(self, attr_str):
        if not set(['entry_id']).issubset(raw_attributes.keys()):
            return None
        match_result = self.check_format(raw_attributes['entry_id'])
        if match_result is None:
            return None
        attr_str = self.cut_string(raw_attributes['entry_id'], attr_str).upper()
        return attr_str
            
    def get_platform(self, raw_attributes):
        """ returns the suitable platform based on the filename """
        platform_map = dict(S1A = 'SENTINEL-1A', S1B = 'SENTINEL-1B')
        platform_str = self.create_attr_string('platform')
        if platform_str is None:
            return None
        return utils.get_gcmd_platform(platform_map[platform_str])
```

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def check_format(self, raw_attributes_entry_id):
+        """ For more information, please check the format of filename to be correct as declared in
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        return re.match('S1\D_\D{2}_\D{4}_\d\D{3}_\d{8}T\d{6}_\d{8}T\d{6}', raw_attributes_entry_id)
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        m = re.match(r'^(?P<platform>S\d[AB])_(?P<mode>[A-Z]{2})_(?P<type>[A-Z]{3})(?P<resolution>[A-Z])_(?P<processing_level>[12])(?P<class>[SA])(?P<polarisationtion>[A-Z]{2})_(?P<time_coverage_start>\d{8}T\d{6})_(?P<time_coverage_end>\d{8}T\d{6})_(?P<orbit>\d{6})_(?P<mission_id>[A-Z0-9]{6})_(?P<product_id>[A-Z0-9]{4})', raw_attributes_entry_id)
+        return m.groupdict()[part] # returns the specific part of the filename per each call of this function
```
aperrin66: There is no need to  have two different regular expressions which match the same thing.
We can also avoid compiling the regex for each method call.

Here would be a way to optimize this a bit:

```python
class SentinelIdentifierMetadataNormalizer(BaseMetadataNormalizer):
    """ Normalizer for extraction of information from the filename """

    MATCHER = re.compile('_'.join([
        r'^(?P<platform>S1[AB])',
        r'(?P<mode>[A-Z]{2})',
        r'(?P<type>[A-Z]{3})(?P<resolution>[A-Z])',
        r'(?P<processing_level>[12])(?P<class>[SA])(?P<polarization>[A-Z]{2})',
        r'(?P<time_coverage_start>\d{8}T\d{6})',
        r'(?P<time_coverage_end>\d{8}T\d{6})',
        r'(?P<orbit>\d{6})',
        r'(?P<mission_id>[A-Z0-9]{6})',
        r'(?P<product_id>[A-Z0-9]{4})',
    ]))

    def check_format(self, raw_attributes_entry_id):
        """ For more information, please check the format of filename to be correct as declared in
        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
        return self.MATCHER.match(raw_attributes_entry_id)

    def cut_string(self, raw_attributes_entry_id, part):
        """ Cuts the filename string based on incoming part """
        return self.MATCHER.match(raw_attributes_entry_id).groupdict()[part]
```

metanorm/normalizers/filename_interpreter.py
```diff
@@ -0,0 +1,103 @@
...
+    def get_entry_id(self, raw_attributes):
+        """ returns the whole raw attribute as the indentifier (or in other words entry_id) """
+        if set(['entry_id']).issubset(raw_attributes.keys()):
+            match_result = self.check_format(raw_attributes['entry_id'])
+            if match_result is not None:
+                return raw_attributes['entry_id']
+            else:
+                return None
+        else:
+            return None
```
akorosov, line 38:
```suggestion
        if set(['Identifier']).issubset(raw_attributes.keys()):
            return raw_attributes['Identifier']
        return None
```

metanorm/normalizers/filename_interpreter.py
aperrin66: One more thing, if this normalizer is just for Sentinel-1 data, it should be indicated in the name of the normalizer class and the name of the file.
Something like `SentinelOneIdentifierMetadataNormalizer` for the class, and a matching name for the file like `sentinel1_identifier.py`.

aperrin66:
One more thing, if this normalizer is just for Sentinel-1 data, it should be indicated in the name of the normalizer class and the name of the file.
Something like `SentinelOneIdentifierMetadataNormalizer` for the class, and a matching name for the file like `sentinel1_identifier.py`.

tests/normalizers/test_filename_interpreter.py
```diff
@@ -0,0 +1,100 @@
...
+                                   "provider"
+                                   ]
+        self.normalizer = normalizers.filename_interpreter.SentinelIdentifierMetadataNormalizer(
+            DATASET_PARAMETER_NAMES)
+
+    def tearDown(self):
+        self.normalizer = None
+
+    def test_none_identification_for_incorrect_filename(self):
+        """ Shall return the correct sentinel based on the filename """
```
akorosov, line 29:
```suggestion
        """ Shall return None from the incorrect sentinel Identifier """
```

aanersc:
All other changes are in 12d2df0. And the normalizer is in 78c0eb2. I think it is ready to merge(78c0eb2) if you are agreed.

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        return self.MATCHER.match(raw_attributes_entry_id).groupdict()[part]
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the whole raw attribute as the indentifier """
+        instrument_str = self.create_attr_string('platform', raw_attributes)
+        if instrument_str is None:
+            return None
```
aperrin66, line 43: What is the purpose of these three lines?

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+        return utils.get_gcmd_platform(platform_map[platform_str])
+
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the filename """
+        instrument_str = self.create_attr_string('platform', raw_attributes)
+        if instrument_str is None:
+            return None
+        # This if is only for safety checking
+        if instrument_str.upper()[:2] == 'S1':
```
aperrin66, line 71: Same here, all this can pretty much be replaced by the `check_format` method.

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+        """ For more information, please check the format of filename to be correct as declared in
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        return self.MATCHER.match(raw_attributes_entry_id)
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        return self.MATCHER.match(raw_attributes_entry_id).groupdict()[part]
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the whole raw attribute as the indentifier """
```
aperrin66, line 40:
```suggestion
        """ returns the whole raw attribute as the identifier """
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+        if instrument_str is None:
+            return None
+        # This if is only for safety checking
+        if instrument_str.upper()[:2] == 'S1':
+            return utils.get_gcmd_instrument('C-SAR')
+
+    def get_time_coverage_start(self, raw_attributes):
+        """ returns the suitable time_coverage_start based on the filename """
+        s_time_str = self.create_attr_string(
+            'time_coverage_start', raw_attributes)
```
aperrin66, line 77:
```suggestion
        s_time_str = self.create_attr_string('time_coverage_start', raw_attributes)
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+        if s_time_str is None:
+            return None
+        return dateutil.parser.parse(s_time_str)
+
+    def get_time_coverage_end(self, raw_attributes):
+        """ returns the suitable time_coverage_end based on the filename """
+        e_time_str = self.create_attr_string(
+            'time_coverage_end', raw_attributes)
```
aperrin66, line 85:
```suggestion
        e_time_str = self.create_attr_string('time_coverage_end', raw_attributes)
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+            return None
+        return dateutil.parser.parse(e_time_str)
+
+    def get_provider(self, raw_attributes):
+        """ returns the suitable provider based on the filename """
+        instrument_str = self.create_attr_string('platform', raw_attributes)
+        if instrument_str is None:
+            return None
+        # This if is only for safety checking
+        if instrument_str.upper()[:2] == 'S1':
```
aperrin66, line 96: Same here, all this can pretty much be replaced by the `check_format` method.

tests/normalizers/test_filename_interpreter.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    def setUp(self):
+        DATASET_PARAMETER_NAMES = ["entry_id",
+                                   "platform",
+                                   "instrument",
+                                   "time_coverage_start",
+                                   "time_coverage_end",
+                                   "provider"
+                                   ]
```
aperrin66, line 22:
```suggestion
        DATASET_PARAMETER_NAMES = [
            "entry_id",
            "platform",
            "instrument",
            "time_coverage_start",
            "time_coverage_end",
            "provider"
        ]
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the whole raw attribute as the indentifier """
+        instrument_str = self.create_attr_string('platform', raw_attributes)
+        if instrument_str is None:
+            return None
+        return raw_attributes['Identifier']
+
+    def create_attr_string(self, attr_str, raw_attributes):
+        """in the case of existance of 'metadata_name' in row data, it will check the format of it
```
aperrin66, line 47:
```suggestion
        """in the case of existence of 'metadata_name' in row data, it will check the format of it
```

aanersc:
@akorosov @aperrin66 
fd52551 is ready to merge, IF YOU ARE AGREE. You can also have a look at 2bbb3cb which is the result of simplification based on Adriens comments.

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+    ]))
+
+    def check_format(self, raw_attributes_entry_id):
+        """ For more information, please check the format of filename to be correct as declared in
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        return self.MATCHER.match(raw_attributes_entry_id)
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        return self.MATCHER.match(raw_attributes_entry_id).groupdict()[part]
```
akorosov, line 38:
```suggestion
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def create_attr_string(self, attr_str, raw_attributes):
+        """in the case of existance of 'self.metadata_name' in row data, it will check the format of it
+        and cut the desired part of the string from the raw attribute of 'self.metadata_name' """
+        if not set([self.metadata_name]).issubset(raw_attributes.keys()):
+            return None
+        match_result = self.check_format(raw_attributes[self.metadata_name])
+        if match_result is None:
+            return None
+        return self.cut_string(raw_attributes[self.metadata_name], attr_str).upper()
```
akorosov, line 53:
```suggestion
    def match_identifier(self, raw_attributes, attr_name='platform'):
        """ Find Identifier in raw_attributes and match agains pattern. Return metadata corresponding to  attr_name """
        if set([self.metadata_name]).issubset(raw_attributes.keys()):
            match_result = self.MATCHER.match(raw_attributes_entry_id)
            if match_result:
                return match_result.groupdict()[attr_name].upper()
        return None
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+        "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions" """
+        return self.MATCHER.match(raw_attributes_entry_id)
+
+    def cut_string(self, raw_attributes_entry_id, part):
+        """ Cuts the filename string based on incoming part """
+        return self.MATCHER.match(raw_attributes_entry_id).groupdict()[part]
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the whole raw attribute as the indentifier """
+        if self.check_format(raw_attributes[self.metadata_name]):
```
akorosov, line 42:
```suggestion
        if self.match_identifier(raw_attributes):
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+            return None
+        match_result = self.check_format(raw_attributes[self.metadata_name])
+        if match_result is None:
+            return None
+        return self.cut_string(raw_attributes[self.metadata_name], attr_str).upper()
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable platform based on the filename """
+        platform_map = dict(S1A='SENTINEL-1A', S1B='SENTINEL-1B')
+        platform_str = self.create_attr_string('platform', raw_attributes)
```
akorosov, line 58:
```suggestion
        platform_str = self.match_identifier(raw_attributes, 'platform')
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+        if platform_str is None:
+            return None
+        return utils.get_gcmd_platform(platform_map[platform_str])
+
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the filename """
+        if self.check_format(raw_attributes[self.metadata_name]):
```
akorosov, line 65:
```suggestion
        if self.match_identifier(raw_attributes):
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+            return utils.get_gcmd_instrument('C-SAR')
+
+    def get_time_coverage_start(self, raw_attributes):
+        """ returns the suitable time_coverage_start based on the filename """
+        s_time_str = self.create_attr_string('time_coverage_start', raw_attributes)
```
akorosov, line 70:
```suggestion
        s_time_str = self.match_identifier(raw_attributes, 'time_coverage_start')
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+        if s_time_str is None:
+            return None
+        return dateutil.parser.parse(s_time_str)
+
+    def get_time_coverage_end(self, raw_attributes):
+        """ returns the suitable time_coverage_end based on the filename """
+        e_time_str = self.create_attr_string('time_coverage_end', raw_attributes)
```
akorosov, line 77:
```suggestion
        e_time_str = self.match_identifier(raw_attributes, 'time_coverage_end')
```

metanorm/normalizers/sentinel1_identifier.py
```diff
@@ -0,0 +1,85 @@
...
+        if e_time_str is None:
+            return None
+        return dateutil.parser.parse(e_time_str)
+
+    def get_provider(self, raw_attributes):
+        """ returns the suitable provider based on the filename """
+        if self.check_format(raw_attributes[self.metadata_name]):
```
akorosov, line 84:
```suggestion
        if self.match_identifier(raw_attributes):
```

metanorm/normalizers/sentinel1_identifier.py
aperrin66: I'm not a fan of this solution, because it becomes unclear that for `get_entry_id()`, `get_instrument()` and `get_provider()`, all that is needed is to check that the `Identifier` matches the expected format.
(Also, in these cases, some useless processing would be done)

Maybe merging `create_attr_string()` and `cut_string()` would be an acceptable compromise?

akorosov: I though about that. The problem is that we cannot call `check_format(raw_attributes[self.metadata_name])` from methods because it will raise an error if self.metadata_name is not in raw_attributes. And error handling is not implemented. So we either have to again check that Identifier is in raw_attributes. So it was simple to create just one function that both checks if Identifier is in metadata and that the identifier matches the pattern.

akorosov: Extracting a value from matched pattern was just an addon. Maybe this addon can be moved to another function.

akorosov: Or to each method...

akorosov: @aperrin66 , better now?
BTW, overhead of match is negligible, i guess...

tests/normalizers/test_sentinel1_identifier.py
```diff
@@ -0,0 +1,102 @@
...
+            "time_coverage_start",
+            "time_coverage_end",
+            "provider"
+        ]
+        self.normalizer = normalizers.sentinel1_identifier.SentinelOneIdentifierMetadataNormalizer(
+            DATASET_PARAMETER_NAMES)
+
+    def tearDown(self):
+        self.normalizer = None
+
```
akorosov, line 29: Also add a test when Identifier is not in input metadata

metanorm/normalizers/sentinel1_identifier.py
akorosov, line 58:
```suggestion
        platform_str = self.match_identifier(raw_attributes).get('platform', None)
```

metanorm/normalizers/sentinel1_identifier.py
aperrin66: Yes, that works.
And the match operation should be really quick since we used the compiled regex.

akorosov:
Looks great! One more thing from previous comments: Also add a test when Identifier is not in input metadata

aanersc:
@akorosov 
@aperrin66 
One more test is add in a1e6dea and covarage reaches 100 percent! The new test is checking the 'Iifi' raw_attribute which is incorrect. It is ready to merge.

Issue #9, opened 2020-07-08
SentinelOneIdentifier should also return Parameters


akorosov:
* To add get_parameters() method which returns a list with one entry:
```python
return [pti.get_wkv_variable('surface_backwards_scattering_coefficient_of_radar_wave')]   
```

akorosov:
Parameter has fields:
standard_name, short_name, units. 

The source of data in the Parameter tabel in geospaas is two pythesint lists:
get_wkv_variable_list
get_cf_standard_name_list

Therefore get_parameter (or get_geospaas_parameter) should return either get_wkv_variable() or get_cf_standard_name() depending on which one succeeds.

aanersc:
@akorosov 
this part of code is ready. The remaining should be modified in the harvesting reop.

aanersc:
the regex should become more generic!

PR #13, opened 2020-07-15
#9 default values for dataset_parameter is set//names are changed in the tests

aanersc:
In order to pass the tests of harvesting this modifications must be made in the normalizer. New test is added to harvesting repo. That test alongside others, require this changes.

### DIFF
--- a/metanorm/normalizers/sentinel_safe.py
+++ b/metanorm/normalizers/sentinel_safe.py
@@ -68,7 +68,7 @@ class SentinelSAFEMetadataNormalizer(BaseMetadataNormalizer):
                 instrument = raw_attributes['Instrument']
 
             additional_keywords = []
-            if set(['Satellite name']).issubset(raw_attributes.keys()) and raw_attributes['Satellite name']!='':
+            if set(['Satellite name']).issubset(raw_attributes.keys()) and raw_attributes['Satellite name']:
                 additional_keywords.append(raw_attributes['Satellite name'])
             return utils.get_gcmd_instrument(instrument, additional_keywords)
         else:
--- a/metanorm/normalizers/sentinel_safe.py
+++ b/metanorm/normalizers/sentinel_safe.py
@@ -67,7 +67,7 @@ class SentinelSAFEMetadataNormalizer(BaseMetadataNormalizer):
                 instrument = raw_attributes['Instrument']
 
             additional_keywords = []
-            if set(['Satellite name']).issubset(raw_attributes.keys()) and raw_attributes['Satellite name']:
+            if set(['Satellite name']).issubset(raw_attributes.keys()):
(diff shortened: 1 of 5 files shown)


aanersc:
@aperrin66 @akorosov 
It is ready now! There are some tiny modifications needed for tests of the new normalizer in order to pass both tests of metanorm and harvesting repo

metanorm/normalizers/sentinel_safe.py
```diff
@@ -61,13 +61,14 @@ def get_instrument(self, raw_attributes):
...
             # This is ugly but pythesint can't find C-SAR from 'SAR-C'
+            #There is a similar code in "setinel1_identifier" normalizer
             if 'SAR-C' in raw_attributes['Instrument']:
                 instrument = 'C-SAR'
             else:
                 instrument = raw_attributes['Instrument']
 
             additional_keywords = []
-            if set(['Satellite name']).issubset(raw_attributes.keys()):
+            if set(['Satellite name']).issubset(raw_attributes.keys()) and raw_attributes['Satellite name']!='':
```
aperrin66, line 71: If you do it the following way, it will also work in case `raw_attributes['Satellite name']` is None
```suggestion
            if set(['Satellite name']).issubset(raw_attributes.keys()) and raw_attributes['Satellite name']:
```
But is this test even necessary? Did you find a case where it is a problem?

metanorm/normalizers/sentinel_safe.py
```diff
@@ -61,13 +61,14 @@ def get_instrument(self, raw_attributes):
         """Get the instrument from the attributes'"""
         if set(['Instrument']).issubset(raw_attributes.keys()):
             # This is ugly but pythesint can't find C-SAR from 'SAR-C'
+            #There is a similar code in "setinel1_identifier" normalizer
```
aperrin66, line 64: Is there? I don't see it.

metanorm/normalizers/sentinel_safe.py
aanersc: You are right, I omit that ending part of the line. It is not necessary.

metanorm/normalizers/sentinel_safe.py
aanersc: This one is also should be removed.

aanersc: Done in de9b10c.

aanersc:
I accidentally make the last commit twice. SO, c13529b and de9b10c are the same. 

To anyway, c13529b is ready if you agreed.

metanorm/normalizers/sentinel_safe.py
aanersc: Done in fcca2cd

Issue #10, opened 2020-07-08
Add normalizers for OSISAF data


akorosov:
This is connected to issue in harvesters: https://github.com/nansencenter/django-geo-spaas-harvesting/issues/16

Normalizers for metadata from ingestors (similar to SentinelSAFEMetadataNormalizer) and from Identifier (similar to SentinelOneIdentifierMetadataNormalizer) should be added.

The identifier normalizer should add Parameters:
standard_name = sea_ice_area_fraction

It is important to keep in mid extendability to add more sources of data and more parameters from OSISAF.

akorosov:
Raw metadata from MET OSISAF:
https://thredds.met.no/thredds/dodsC/osisaf/met.no/ice/amsr2_conc/2020/07/ice_conc_nh_polstere-100_amsr2_202007121200.nc.ddx

akorosov:
After sea ice concentration normalizers for sea ice type and sea ice drift should be added
Parameters:
sea_ice_x_displacement
sea_ice_y_displacement

sea_ice_classification

aanersc:
Everything needed is written in a3466b7 except parameter assignment. I need the start name of parameters for it. Please provide them for me @akorosov .

akorosov:
@aanersc , I already wrote these names. See above.

aanersc:
@aperrin66 
Please have a look at this new normalizer, If you agree with in general, then I start to write some tests for it. Find it in 7849bf9 please.

aanersc:
Dear @akorosov,

PLease check out the new normalizer of OSISAF from remote sensing point of view in order to deliver meaningfull results in 'get_platform' and 'get_instrument' methods of it. It is necessary to mention this aspect of work.

akorosov:
For all products you can use:
pti.get_gcmd_platform('Earth Observation Satellites')     

And these instruments:

https://thredds.met.no/thredds/catalog/osisaf/met.no/ice/conc/catalog.html
pti.get_gcmd_instrument('Imaging Spectrometers/Radiometers')    

https://thredds.met.no/thredds/catalog/osisaf/met.no/ice/amsr2_conc/catalog.html
pti.get_gcmd_instrument('AMSR2') 

https://thredds.met.no/thredds/catalog/osisaf/met.no/ice/drift_lr/merged/catalog.html
pti.get_gcmd_instrument('Earth Remote Sensing Instruments')

[...]

PR #14, opened 2020-07-23
Issue10 osisaf reading and para assign

aanersc:


metanorm/normalizers/base.py
```diff
@@ -84,22 +81,23 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
...
                 if callable(param_method):
-                    harv_parameters[param] = param_method(raw_attributes)
+                    output_parameter_names[param] = param_method(raw_attributes)
                 else:
                     raise MetadataNormalizationError(
                         f"Unable to find a value for the {param} parameter")
-
-        return harv_parameters
+        #adding (updating the normalized harvested parameters with the repetitive ones)
+        output_parameter_names.update(output_cumulative_parameter_names)
```
aanersc, line 102: @aperrin66 
I move this line here because of recursive nature of calling the normalize function. In this way the update is only done once per each raw attribute not per each normalizer of that raw attribute.

aanersc:
@aperrin66 
The old tests are modified slightly in order to pass with new design in ff86bfc.

aanersc:
@aperrin66 
I have finalized the tests by adding a new one based on the new design of accumulative ones (in this case, `dataset_parameters`) in this commit c402206. Now it is ready to merge.

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
+        handler.NORMALIZERS = [
+        normalizers.OSISAFMetadataNormalizer,
+        normalizers.SentinelSAFEMetadataNormalizer,
+        normalizers.SentinelOneIdentifierMetadataNormalizer,
+        normalizers.ACDDMetadataNormalizer,
+        normalizers.GeoSpatialWellKnownMetadataNormalizer,
+        normalizers.GeoSpatialDefaultMetadataNormalizer,
+        ]
+        n = handler._chain
+        result=n.normalize({'Identifier': 'S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1', 'product_name':'osi_saf_mr_ice_drift'})
```
aperrin66, line 49: why not just use `handler.get_parameters(...)`?

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
+        self.assertEqual(len(result['dataset_parameters']),3)
+        self.assertIn('sea_ice_y_displacement',[i['standard_name'] for i in result['dataset_parameters']])
+        self.assertIn('sea_ice_x_displacement',[i['standard_name'] for i in result['dataset_parameters']])
+        self.assertIn('surface_backwards_scattering_coefficient_of_radar_wave',[i['standard_name'] for i in result['dataset_parameters']])
```
aperrin66, line 53: Why not just compare with the expected output dictionary? It would be much easier to understand.

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
+        """Shall return consistent results for cumulative parameters
+        even if the ordering of normalizers are changed"""
+        handler = handlers.GeospatialMetadataHandler(['platform'],['dataset_parameters'])
+        handler.NORMALIZERS = [
+        normalizers.OSISAFMetadataNormalizer,
+        normalizers.SentinelSAFEMetadataNormalizer,
+        normalizers.SentinelOneIdentifierMetadataNormalizer,
+        normalizers.ACDDMetadataNormalizer,
+        normalizers.GeoSpatialWellKnownMetadataNormalizer,
+        normalizers.GeoSpatialDefaultMetadataNormalizer,
```
aperrin66, line 46: please indent this properly

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
+        self.assertIn('surface_backwards_scattering_coefficient_of_radar_wave',[i['standard_name'] for i in result['dataset_parameters']])
+        #Redefinition with another order of normalizers
+        handler = handlers.GeospatialMetadataHandler(['platform'],['dataset_parameters'])
+        handler.NORMALIZERS = [
+        normalizers.SentinelSAFEMetadataNormalizer,
+        normalizers.SentinelOneIdentifierMetadataNormalizer,
+        normalizers.ACDDMetadataNormalizer,
+        normalizers.OSISAFMetadataNormalizer,
+        normalizers.GeoSpatialWellKnownMetadataNormalizer,
+        normalizers.GeoSpatialDefaultMetadataNormalizer,
```
aperrin66, line 62: please indent this properly

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
         _ = handler.get_parameters({})
         mock_normalize.assert_called()
+
+class OrderingOfNormalizer(unittest.TestCase):
+    """Tests for the ordering of normalizers. """
+
+    def test_dataset_parameters(self):
+        """Shall return consistent results for cumulative parameters
+        even if the ordering of normalizers are changed"""
+        handler = handlers.GeospatialMetadataHandler(['platform'],['dataset_parameters'])
```
aperrin66, line 39: No need to use `GeospatialMetadataHandler` if you are going to override the `NORMALIZERS` attribute.
Just use `MetadataHandler`

tests/test_handler.py
aperrin66, line 46: Why use that many normalizers? Wouldn't `OSISAFMetadataNormalizer` and `SentinelOneIdentifierMetadataNormalizer` be enough?

tests/test_handler.py
```diff
@@ -13,19 +13,57 @@ class GeospatialMetadataHandlerTestCase(unittest.TestCase):
...
         """Checks that the get_parameters calls the BaseMetadaNormalizer.normalize method"""
-        handler = handlers.GeospatialMetadataHandler(['test_parameter'])
+        handler = handlers.GeospatialMetadataHandler(['test_output_parameter_names'],['test_cumulative_parameter_names'])
         _ = handler.get_parameters({})
         mock_normalize.assert_called()
+
+class OrderingOfNormalizer(unittest.TestCase):
+    """Tests for the ordering of normalizers. """
+
+    def test_dataset_parameters(self):
```
aperrin66, line 36: - the name should reflect the purpose of the test, so something about normalizers order
- I am not sure this is the right place for that test. It concerns the behavior of normalizers and mostly the `normalize()` method, so it should probably be put in the `BaseMetadataNormalizerTestCase` in `tests/normalizers/test_base.py`.

aanersc:
@aperrin66 
All of your comments are addressed in 4726823.

tests/test_handler.py
aanersc: Done 4726823

tests/test_handler.py
aanersc: Done 4726823

tests/test_handler.py
aanersc: Done 4726823

tests/test_handler.py
aanersc: Done 4726823

tests/test_handler.py
aanersc: Done 4726823

metanorm/normalizers/base.py
```diff
@@ -25,35 +26,52 @@ def next(self):
     def next(self, normalizer):
         self._next = normalizer
 
-    def normalize(self, raw_attributes, parameters=None):
+    def normalize(self, raw_attributes, output_parameter_names=None, output_cumulative_parameter_names=None):
```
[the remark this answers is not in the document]
aanersc: Done 4726823

metanorm/normalizers/base.py
```diff
@@ -25,35 +26,52 @@ def next(self):
...
         """
-        Loops through parameters and checks if a method is available for each parameter.
+        Loops through output_parameter_names and checks if a method is available for each parameter.
         If so try it out. The 'raw_attributes' dictionary is then passed to the next normalizer, so
         that it can attempt to fill in any None value which might be left.
         """
-        if not parameters:
-            parameters = dict({(key, None) for key in self._parameter_names})
+        if not output_parameter_names:
+            output_parameter_names = dict({(key, None) for key in self._output_parameter_names})
```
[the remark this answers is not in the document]
aanersc: Done 4726823

metanorm/normalizers/base.py
```diff
@@ -25,35 +26,52 @@ def next(self):
     def next(self, normalizer):
         self._next = normalizer
 
-    def normalize(self, raw_attributes, parameters=None):
+    def normalize(self, raw_attributes, output_parameter_names=None, output_cumulative_parameter_names=None):
         """
-        Loops through parameters and checks if a method is available for each parameter.
+        Loops through output_parameter_names and checks if a method is available for each parameter.
```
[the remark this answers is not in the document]
aanersc: Done 4726823

metanorm/normalizers/base.py
```diff
@@ -25,35 +26,52 @@ def next(self):
...
+                if output_parameter_names[param]:
                     LOGGER.debug("%s: found a value for the '%s' parameter",
                                  self.__class__.__name__, param)
                 else:
                     LOGGER.debug("%s: value not found for the '%s' parameter",
                                  self.__class__.__name__, param)
 
+        for param in output_cumulative_parameter_names.keys():
+            try:
+                if getattr(self, 'get_' + param)(raw_attributes) is not None:
```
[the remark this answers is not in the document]
aanersc: Done 4726823

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+            if raw_attributes['product_name'][-4:] == 'conc':
+                return [pti.get_wkv_variable('sea_ice_area_fraction'), ]
+            elif raw_attributes['product_name'][-5:] == 'drift':
+                return [pti.get_cf_standard_name('sea_ice_x_displacement'),
+                        pti.get_cf_standard_name('sea_ice_y_displacement'), ]
+            elif raw_attributes['product_name'][-4:] == 'type':
+                return [pti.get_cf_standard_name('sea_ice_classification'), ]
+
+        else:
+            return None
```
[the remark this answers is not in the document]
aanersc: Done 4726823

metanorm/normalizers/base.py
```diff
@@ -63,22 +81,23 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
     MetadataNormalizationError. It is strongly advised to return hard coded values in those methods.
     """
 
-    def normalize(self, raw_attributes, parameters=None):
+    def normalize(self, raw_attributes, output_parameter_names=None, output_cumulative_parameter_names=None):
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/base.py
```diff
@@ -63,22 +81,23 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
...
         """
-        Loops through parameters and checks if a method is available for that parameter.
+        Loops through output_parameter_names and checks if a method is available for that parameter.
         If so use it,
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+import metanorm.utils as utils
+
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Test class for osisaf normalizer """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the 'instrument_type' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['instrument_type']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if raw_attributes['product_name'][:7] == 'osi_saf':
+                return utils.UNKNOWN # make it unknown only for osisaf products only
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Test class for osisaf normalizer """
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the 'instrument_type' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['instrument_type']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if raw_attributes['product_name'][:7] == 'osi_saf':
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+            return None
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable instrument based on the 'platform_name' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['platform_name']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_platform(raw_attributes['platform_name'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if raw_attributes['product_name'][:7] == 'osi_saf':
+                return utils.UNKNOWN # make it unknown only for osisaf products only
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_dataset_parameters(self, raw_attributes):
+        """ returns the suitable instrument based on the lasting letters of 'product_name' attribute """
+        if set(['product_name']).issubset(raw_attributes.keys()):
+            if raw_attributes['product_name'][-4:] == 'conc':
+                return [pti.get_wkv_variable('sea_ice_area_fraction'), ]
+            elif raw_attributes['product_name'][-5:] == 'drift':
+                return [pti.get_cf_standard_name('sea_ice_x_displacement'),
+                        pti.get_cf_standard_name('sea_ice_y_displacement'), ]
+            elif raw_attributes['product_name'][-4:] == 'type':
+                return [pti.get_cf_standard_name('sea_ice_classification'), ]
```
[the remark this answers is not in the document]
aanersc: @aperrin66 
You are right. It  should be available in the raw attributes.**In practice**, there is no difference to read it from NC_GLOBAL or read it somewhere else(**No more information of metadata is added**). For example here is to read it from somewhere else in the raw attributes:
image


And here is the way that we read it from product_name:
image

There is no advantage of read it from somewhere else. We need add some extra code in order to search and find it outside of NC_GLOBAL. Is it worthwhile to put extra code for it? I think the easiest and efficient solution is to use the product_name from NC_GLOBAL which works for all three of them (comc, type and drift)

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_platform(self):
+        """platform from OSISAFMetadataNormalizer"""
+        attributes = {'platform_name': 'value_1', 'activity_type': 'value_2'}
+        # 'instrument_type' must priortized in this normalizer
+        self.assertEqual(self.normalizer.get_platform(attributes)['Short_Name'], 'value_1')
+
+        attributes = {'activity_type': 'value_2'}
+        # in the absence of 'instrument_type' value should be None regardless of 'activity_type' attribute
+        self.assertIsNone(self.normalizer.get_instrument(attributes))
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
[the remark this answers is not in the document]
aanersc: Done in the next commit

tests/test_handler.py
aanersc: @aperrin66 
If you mean get_dataset_parameters(....), it is not the objective of this test. This tests want to see that wherher parameters can be added from different sources or not (regardless of ordering of normalizers).

If I put get_parameters(...) over there, here is the result:
image

metanorm/normalizers/base.py
```diff
@@ -25,35 +26,52 @@ def next(self):
...
+            if output_cumulative_parameter_names[param]:
+                LOGGER.debug("%s: found a value for the '%s' parameter",
+                             self.__class__.__name__, param)
+            else:
+                LOGGER.debug("%s: value not found for the '%s' parameter",
+                             self.__class__.__name__, param)
+
         if self.next is None:
-            return parameters
+            return output_parameter_names
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            return None
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable instrument based on the 'platform_name' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['platform_name']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_platform(raw_attributes['platform_name'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if raw_attributes['product_name'][:7] == 'osi_saf':
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

tests/test_handler.py
aanersc: Done in the next commit

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,97 @@
+"""Tests for the OSISAF metadata normalizer"""
+import unittest
+from collections import OrderedDict
+
+import dateutil
+from dateutil.tz import tzutc
+from django.contrib.gis.geos.geometry import GEOSGeometry
+import metanorm.handlers as handlers
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

tests/normalizers/test_osisaf.py
```diff
@@ -53,11 +74,11 @@ def test_dataset_parameters(self):
...
-                                      ('long_name', 'Sea Ice Concentration'),
-                                      ('short_name', 'ice_conc'),
-                                      ('units', '%'),
-                                      ('minmax', '0 100'),
-                                      ('colormap', 'jet')]))
+                                      ('canonical_units', '1'),
+                                      ('grib', '91'),
+                                      ('amip', 'sic'),
+                                      ('description', '"X_area_fraction" means the fraction of horizontal area occupied by X. "X_area" means the horizontal area occupied by X within the grid cell. Sea ice area fraction is area of the sea surface occupied by sea ice. It is also called "sea ice concentration".'),
+                                      ]))
```
aanersc: This is related to this issue:https://github.com/nansencenter/metanorm/issues/15#issue-668611590

tests/normalizers/test_base.py
```diff
@@ -104,20 +106,85 @@ def test_normalize_parameter_with_existing_method(self):
...
+        self.assertIn(OrderedDict([('standard_name','sea_ice_x_displacement'),
+            ('canonical_units','m'),
+            ('grib',''),
+            ('amip',''),
+            ('description','"x" indicates a vector component along the grid x-axis, positive with increasing x. "Displacement" means the change in geospatial position of an object that has moved over time. If possible, the time interval over which the motion took place should be specified using a bounds variable for the time coordinate variable.  A displacement can be represented as a vector. Such a vector should however not be interpreted as describing a rectilinear, constant speed motion but merely as an indication that the start point of the vector is found at the tip of the vector after the time interval associated with the displacement variable.  A displacement does not prescribe a trajectory. Sea ice displacement can be defined as a two-dimensional vector, with no vertical component. An x displacement is calculated from the difference in the moving object\'s grid x coordinate between the start and end of the time interval associated with the displacement variable.')]),result['dataset_parameters'])
```
aperrin66, line 147: you can mock the call to pythesint to avoid having this lengthy data in the test

metanorm/handlers.py
```diff
@@ -20,16 +20,19 @@ class MetadataHandler():
     # THE LAST NORMALISER MUST INHERIT FROM BaseDefaultMetadataNormalizer
     NORMALIZERS = []
 
-    def __init__(self, parameter_names):
+    def __init__(self, output_parameter_names, output_cumulative_parameter_names):
         """Builds a chain of normalizers for the given parameter names"""
+        if normalizers.geospatial_defaults.GeoSpatialDefaultMetadataNormalizer not in self.NORMALIZERS:
+            raise ValueError("GeoSpatialDefaultMetadataNormalizer must be in the list of normalizers")
```
aperrin66, line 26: - That does not test that the default normalizer is last, just that it is in the list
- You shouldn't test for a specific default normalizer, but that the last normalizer in the list is a (class that inherits from) `BaseDefaultMetadataNormalizer`. Use `isinstance()` for this.

metanorm/handlers.py
```diff
@@ -20,16 +20,19 @@ class MetadataHandler():
...
-        self._chain = last_normalizer = self.NORMALIZERS[0](parameter_names)
+        self._chain = last_normalizer = self.NORMALIZERS[0](output_parameter_names, output_cumulative_parameter_names)
 
         for normalizer_class in self.NORMALIZERS[1:]:
-            current_normalizer = normalizer_class(parameter_names)
+            current_normalizer = normalizer_class(output_parameter_names, output_cumulative_parameter_names)
             last_normalizer.next = current_normalizer
             last_normalizer = current_normalizer
 
+
```
aperrin66, line 35:
```suggestion
```

metanorm/handlers.py
```diff
@@ -38,9 +41,12 @@ def get_parameters(self, raw_attributes):
 class GeospatialMetadataHandler(MetadataHandler):
     """Geospatial metadata handler"""
     NORMALIZERS = [
+        normalizers.NETCDFMetadataNormalizer,
+
```
aperrin66, line 45:
```suggestion
```

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,97 @@
+"""Tests for the OSISAF metadata normalizer"""
```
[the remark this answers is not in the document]
aperrin66: That is still valid

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,113 @@
...
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Container class of get_ methods for osisaf normalizer """
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the 'instrument_type' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['instrument_type']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                return utils.get_gcmd_instrument('UNKNOWN') # make it unknown for osisaf products only
```
aperrin66, line 27: please use `utils.UNKNOWN`

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,113 @@
...
+            return None
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable instrument based on the 'platform_name' attribute (priortized one)
+        and 'activity_type' attribute """
+        if set(['platform_name']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_platform(raw_attributes['platform_name'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                return utils.get_gcmd_platform('UNKNOWN') # make it unknown for osisaf products only
```
aperrin66, line 38: please use `utils.UNKNOWN`

tests/test_handler.py
```diff
@@ -10,22 +10,28 @@
 
 class GeospatialMetadataHandlerTestCase(unittest.TestCase):
     """Test the GeospatialMetadataHandler class"""
+    def test_build_normalizer_chain(self):
+        """It is mandatory to have 'GeoSpatialDefaultMetadataNormalizer' in the list of normalizers.
```
aperrin66, line 14: No, it's mandatory to have a class inheriting from `BaseDefaultMetadataNormalizer` as the last normalizer in the list.
For `GeospatialMetadataHandler`, it is indeed `GeoSpatialDefaultMetadataNormalizer` but it would be different for other handlers

tests/test_handler.py
```diff
@@ -10,22 +10,28 @@
...
+        This normalizer should never be removed from the list of normalizers"""
+        handler = handlers.GeospatialMetadataHandler(['test_output_parameters'],['test_repetitive_parameter'])
+
+        self.assertTrue(normalizers.geospatial_defaults.GeoSpatialDefaultMetadataNormalizer in handler.NORMALIZERS)
```
aperrin66, line 18: it should be the last in the list, not just in it

tests/test_handler.py
```diff
@@ -10,22 +10,28 @@
...
 
     def test_build_normalizer_chain(self):
         """Instantiate a handler and check the normalizer chain is correctly built"""
-        handler = handlers.GeospatialMetadataHandler(['test_parameter'])
+        handler = handlers.GeospatialMetadataHandler(['test_output_parameters'],['test_repetitive_parameter'])
```
aperrin66, line 22:
```suggestion
        handler = handlers.GeospatialMetadataHandler(['test_output_parameter'],['test_repetitive_parameter'])
```

tests/test_handler.py
```diff
@@ -10,22 +10,28 @@
...
 
         normalizer = handler._chain
         i = 0
         while normalizer:
             self.assertIsInstance(normalizer, handler.NORMALIZERS[i])
-            self.assertEqual(normalizer._parameter_names, ['test_parameter'])
+            self.assertEqual(normalizer._output_parameters, ['test_output_parameters'])
```
aperrin66, line 28:
```suggestion
            self.assertEqual(normalizer._output_parameters, ['test_output_parameter'])
```

aperrin66, line 28: please also test that the cumulative parameter names are correctly assigned

tests/test_handler.py
```diff
@@ -10,22 +10,28 @@
...
             normalizer = normalizer.next
             i += 1
 
     @mock.patch.object(normalizers.BaseMetadataNormalizer, 'normalize', return_value={})
     def test_get_parameters_calls_normalize(self, mock_normalize):
         """Checks that the get_parameters calls the BaseMetadaNormalizer.normalize method"""
-        handler = handlers.GeospatialMetadataHandler(['test_parameter'])
+        handler = handlers.GeospatialMetadataHandler(['test_output_parameters'],['test_cumulative_parameter_names'])
```
aperrin66, line 35:
```suggestion
        handler = handlers.GeospatialMetadataHandler(['test_output_parameter_name'],['test_cumulative_parameter_name'])
```

metanorm/normalizers/NETCDF_CF.py
```diff
@@ -0,0 +1,24 @@
...
+
+import pythesint as pti
+
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class NETCDFMetadataNormalizer(BaseMetadataNormalizer):
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

I think **NETCDFCFMetadataNormalizer** is a better choice since **CFMetadataNormalizer** is vague and does not provide anything for us.

metanorm/normalizers/NETCDF_CF.py
```diff
@@ -0,0 +1,24 @@
...
+    """Generate the GeoSPaaS Datasetparameters using raw_dataset_parameters attribute"""
+
+    def get_dataset_parameters(self, raw_attributes):
+        """ Get the dataset's parameters """
+        if set(['raw_dataset_parameters']).issubset(raw_attributes.keys()):
+            standardized_dataset_parameters = list()
+            for raw_parameter_name in raw_attributes['raw_dataset_parameters']:
+                if len(pti.search_cf_standard_name_list(raw_parameter_name)) > 0:
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

metanorm/normalizers/base.py
```diff
@@ -11,9 +11,21 @@
 class BaseMetadataNormalizer():
     """Base class for standard normalizers"""
 
-    def __init__(self, parameter_names):
-        """parameter_names: iterable"""
-        self._parameter_names = parameter_names
+    def __init__(self, output_parameters=None, output_cumulative_parameters=None):
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

As far as I remember, previous time, we changed it based on your suggestion to these ones and now switch back again!!!  :) 

Actually the base.py is executed inside the normalize method.

metanorm/normalizers/base.py
```diff
@@ -11,9 +11,21 @@
...
+        """
+        output_parameter_names and output_cumulative_parameter_names should contain the lists of
+        desired parameters names to extract.
+        Standard parameter are filled by the first normalizer which find a value for them.
+        Cumulative parameters are lists to which each normalizer appends the values it finds.
+        """
+        self._output_parameters = output_parameters
+        self._output_cumulative_parameters = output_cumulative_parameters
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -11,9 +11,21 @@
...
 
-    def __init__(self, parameter_names):
-        """parameter_names: iterable"""
-        self._parameter_names = parameter_names
+    def __init__(self, output_parameters=None, output_cumulative_parameters=None):
+        """
+        output_parameter_names and output_cumulative_parameter_names should contain the lists of
+        desired parameters names to extract.
+        Standard parameter are filled by the first normalizer which find a value for them.
+        Cumulative parameters are lists to which each normalizer appends the values it finds.
```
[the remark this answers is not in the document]
aanersc: In handlers.py there is no mentioning or no usage of them. I think this is the right place to talk about _Standard parameter_ and _Cumulative parameters_ in the docstring because they are declared and used here in this base.py file.

From the **workflow point of view**, you are right. But if we want to move them to the handlers.py then we have to refer the code reader to read the dosctring of handlers.py as well.

Your kind feedback?

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aperrin66: Last time I was talking about the arguments of the `normalize()` method. Now it is about the arguments of the constructor.

metanorm/normalizers/base.py
```diff
@@ -25,35 +37,64 @@ def next(self):
     def next(self, normalizer):
         self._next = normalizer
 
-    def normalize(self, raw_attributes, parameters=None):
+    def normalize(self, raw_attributes, output_parameters=None, output_cumulative_parameters=None):
         """
-        Loops through parameters and checks if a method is available for each parameter.
+        Loops through the desired output parameter names and checks if a method is available for each parameter.cumulative parameters are lists to which each normalizer adds the values it finds
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aperrin66: I am talking about the `normalize()` method of the `BaseMetadataNormalizer` class, handlers.py has nothing to do with it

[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +37,64 @@ def next(self):
...
-                if parameters[param]:
+                if output_parameters[param]:
                     LOGGER.debug("%s: found a value for the '%s' parameter",
                                  self.__class__.__name__, param)
                 else:
                     LOGGER.debug("%s: value not found for the '%s' parameter",
                                  self.__class__.__name__, param)
 
+        for param in output_cumulative_parameters.keys():
+            previous_len=len(output_cumulative_parameters[param])
```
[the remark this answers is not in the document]
aanersc: I normally use pylint and formatting function of vs code. however, sometimes one or two file is forgotten to be formatted.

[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +37,64 @@ def next(self):
...
+            try:
+                new_member=getattr(self, 'get_' + param)(raw_attributes)
+                if new_member not in output_cumulative_parameters[param]:
+                    output_cumulative_parameters[param].extend(new_member)
+            except AttributeError:
+                LOGGER.debug("%s: no method available for the '%s' parameter",
+                             self.__class__.__name__, param)
+            except TypeError:
+                LOGGER.debug("%s: no method available for the '%s' parameter",
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +37,64 @@ def next(self):
...
+                LOGGER.debug("%s: value not found for the '%s' parameter",
+                             self.__class__.__name__, param)
+
         if self.next is None:
-            return parameters
+            # the second one that is written for return statement is only for testing purposes, it have no role in the realtime performance of the code
+            return {**output_parameters, **output_cumulative_parameters}
         else:
-            return self.next.normalize(raw_attributes, parameters)
+            #output_cumulative_parameters["dataset_parameters"]=set(output_cumulative_parameters["dataset_parameters"])
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +37,64 @@ def next(self):
...
+            if previous_len<len(output_cumulative_parameters[param]):
+                LOGGER.debug("%s: found a value for the '%s' parameter",
+                             self.__class__.__name__, param)
+            else:
+                LOGGER.debug("%s: value not found for the '%s' parameter",
+                             self.__class__.__name__, param)
+
         if self.next is None:
-            return parameters
+            # the second one that is written for return statement is only for testing purposes, it have no role in the realtime performance of the code
```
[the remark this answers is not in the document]
aanersc: Maybe I use a bad terminology. My mean of performance is the execution of the code in the real case in a real harvesting process. So, it is good to convert that word into execution. This way that sentence become like this:
```suggestion
            # the second one that is written for return statement is only for testing purposes, it have no role in the realtime execution of the code
```

Is it ok?

metanorm/normalizers/base.py
```diff
@@ -63,22 +104,23 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
...
                 param_method = getattr(self, 'get_' + param, None)
                 if callable(param_method):
-                    parameters[param] = param_method(raw_attributes)
+                    output_parameters[param] = param_method(raw_attributes)
                 else:
                     raise MetadataNormalizationError(
                         f"Unable to find a value for the {param} parameter")
-
-        return parameters
+        #adding (updating the normalized harvested parameters with the cumulative ones)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_base.py
```diff
@@ -104,20 +106,85 @@ def test_normalize_parameter_with_existing_method(self):
...
         }
 
-        normalizer = self.TestDefaultMetadataNormalizer(parameter_names)
+        normalizer = self.TestDefaultMetadataNormalizer(output_parameter_names, [])
 
         with self.assertRaises(MetadataNormalizationError):
             _ = normalizer.normalize(attributes)
+
+class OrderingOfNormalizer(unittest.TestCase):
+    """Tests for the ordering of normalizers. """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aperrin66: Then I would say `in a standard execution of the code` rather than `in the realtime execution of the code`.
"real time" means something else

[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_base.py
aanersc: It does not harm anything and the issue of #15 is still related to this one. I dont what to mock and the if another decision is made, then I will forced to change all of those mocks. lets go with these ones and then we may change them if necessary. OK?

aperrin66: sure

metanorm/handlers.py
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: test_output_parameters

aanersc: Done in the next commit.

metanorm/handlers.py
aanersc: Done in the next commit.

metanorm/handlers.py
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,113 @@
...
+            if raw_attributes['product_name'][-8:] == 'ice_conc' and raw_attributes['product_name'][:7] == 'osi_saf':
+                return [pti.get_cf_standard_name('sea_ice_area_fraction'), ]
+            elif raw_attributes['product_name'][-9:] == 'ice_drift' and raw_attributes['product_name'][:7] == 'osi_saf':
+                return [pti.get_cf_standard_name('sea_ice_x_displacement'),
+                        pti.get_cf_standard_name('sea_ice_y_displacement'), ]
+            elif raw_attributes['product_name'][-8:] == 'ice_type' and raw_attributes['product_name'][:7] == 'osi_saf':
+                return [pti.get_cf_standard_name('sea_ice_classification'), ]
+
+        else:
+            return []
```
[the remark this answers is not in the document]
aanersc: based on #15  we decide what to do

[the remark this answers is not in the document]
aanersc: Done in the next commit. It is completely removed here and there is only one in the netcdfcf normalizer.

aanersc:
@aperrin66 
All your comments are addressed except this one: https://github.com/nansencenter/metanorm/pull/14#discussion_r462989172

It needs some time to write tests for them. It is good the receive your feedback about the development aspect of this PR which is completely available in 3861a19.

aanersc:
@aperrin66
The tests for increasing the coverage are added to d5c04d8.

If you agree, it is ready to merge.

tests/test_handler.py
```diff
@@ -11,21 +11,45 @@
...
 
         normalizer = handler._chain
         i = 0
         while normalizer:
             self.assertIsInstance(normalizer, handler.NORMALIZERS[i])
-            self.assertEqual(normalizer._parameter_names, ['test_parameter'])
+            self.assertEqual(normalizer._output_parameters_names, [
+                             'test_output_parameter'])
+            self.assertEqual(normalizer._output_cumulative_parameters_names, [
+                             'test_cumulative_parameter'])
```
aperrin66, line 45:
```suggestion
            self.assertEqual(normalizer._output_parameters_names, ['test_output_parameter'])
            self.assertEqual(normalizer._output_cumulative_parameters_names,
                             ['test_cumulative_parameter'])
```

metanorm/normalizers/NETCDF_CF.py
```diff
@@ -0,0 +1,24 @@
...
+import pythesint as pti
+
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class NETCDFCFMetadataNormalizer(BaseMetadataNormalizer):
+    """Generate the GeoSPaaS Datasetparameters using raw_dataset_parameters attribute"""
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/handlers.py
```diff
@@ -20,13 +20,18 @@ class MetadataHandler():
     # THE LAST NORMALISER MUST INHERIT FROM BaseDefaultMetadataNormalizer
     NORMALIZERS = []
 
-    def __init__(self, parameter_names):
+    def __init__(self, output_parameter_names, output_cumulative_parameter_names):
         """Builds a chain of normalizers for the given parameter names"""
+        if not isinstance(self.NORMALIZERS[-1]([], [],), normalizers.base.BaseDefaultMetadataNormalizer):
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -11,9 +11,19 @@
...
+        """
+        output_parameter_names and output_cumulative_parameter_names should contain the lists of
+        desired parameters names to extract.
+        """
+        self._output_parameters_names = output_parameters_names
+        self._output_cumulative_parameters_names = output_cumulative_parameters_names
+        if self._output_parameters_names is None and self._output_cumulative_parameters_names is None:
+            raise ValueError((
+                "Either output_parameter_names or output_cumulative_parameter_names "
+                "must be specified and not empty"
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +35,66 @@ def next(self):
...
+    def normalize(self, raw_attributes, output_parameters=None, output_cumulative_parameters=None):
         """
-        Loops through parameters and checks if a method is available for each parameter.
+        Loops through the desired output parameter names and checks if a method is available for each parameter.
+        Cumulative parameters are lists to which each normalizer adds the values it finds regardless of normalizers ordering.
+        The search for values of other parameters are stopped as soon as a value is found for it. In this way, ordering
+        of normalizers matter for them (not for cumulative ones).
+
         If so try it out. The 'raw_attributes' dictionary is then passed to the next normalizer, so
         that it can attempt to fill in any None value which might be left.
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +35,66 @@ def next(self):
...
+            output_cumulative_parameters = {key: []
+                                            for key in self._output_cumulative_parameters_names}
+
+        for param in output_parameters.keys():
+            if output_parameters[param] is None:
                 try:
                     # Maybe there is a cleaner way to implement this?
-                    parameters[param] = getattr(self, 'get_' + param)(raw_attributes)
+                    output_parameters[param] = getattr(
+                        self, 'get_' + param)(raw_attributes)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -25,35 +35,66 @@ def next(self):
...
                                  self.__class__.__name__, param)
                 else:
                     LOGGER.debug("%s: value not found for the '%s' parameter",
                                  self.__class__.__name__, param)
 
+        for param in output_cumulative_parameters.keys():
+            previous_len = len(output_cumulative_parameters[param])
+            try:
+                new_member = getattr(self, 'get_' + param)(raw_attributes)
+                if new_member not in output_cumulative_parameters[param]:
```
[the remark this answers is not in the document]
aanersc: Duplicates are also related https://github.com/nansencenter/metanorm/issues/15#issue-668611590. After I have changed the function of pti that has been used by various normalizers to unify and only use `search_cf_standard_name_list` function. They wont produce duplicates!

Here in sentinel1_identifier we use `get_wkv_variable` function which is different from `search_cf_standard_name_list` and this cause the problem. Because this problem must be solved from upstream (from the pti repo).

metanorm/normalizers/base.py
```diff
@@ -25,35 +35,66 @@ def next(self):
...
+            if previous_len < len(output_cumulative_parameters[param]):
+                LOGGER.debug("%s: found a value for the '%s' parameter",
+                             self.__class__.__name__, param)
+            else:
+                LOGGER.debug("%s: value not found for the '%s' parameter",
+                             self.__class__.__name__, param)
+
         if self.next is None:
-            return parameters
+            # the second one that is written for return statement is only for testing purposes, it have no role in a standard execution of the code
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
```diff
@@ -63,22 +104,24 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
...
                 param_method = getattr(self, 'get_' + param, None)
                 if callable(param_method):
-                    parameters[param] = param_method(raw_attributes)
+                    output_parameters[param] = param_method(raw_attributes)
                 else:
                     raise MetadataNormalizationError(
                         f"Unable to find a value for the {param} parameter")
-
-        return parameters
+        # updating the normalized harvested parameters with the cumulative ones
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/NETCDF_CF.py
```diff
@@ -0,0 +1,24 @@
+"""Normalizer for the CF Standard convention"""
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+import metanorm.utils as utils
+
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Container class of get_ methods for osisaf normalizer """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                # make it unknown for osisaf products only
+                return utils.get_gcmd_instrument(utils.UNKNOWN)
+        else:
+            return None
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable instrument based on the 'platform_name' attribute (priortized one)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+            return utils.get_gcmd_platform(raw_attributes['platform_name'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                # make it unknown for osisaf products only
+                return utils.get_gcmd_platform(utils.UNKNOWN)
+        else:
+            return None
+
+    def get_time_coverage_start(self, raw_attributes):
+        """ returns the suitable instrument based on the 'start_date' attribute """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+        if set(['start_date']).issubset(raw_attributes.keys()):
+            return dateutil.parser.parse(raw_attributes['start_date']).replace(tzinfo=tzutc())
+        else:
+            return None
+
+    def get_time_coverage_end(self, raw_attributes):
+        """ returns the suitable instrument based on the 'stop_date' attribute """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+        if set(['stop_date']).issubset(raw_attributes.keys()):
+            return dateutil.parser.parse(raw_attributes['stop_date']).replace(tzinfo=tzutc())
+        else:
+            return None
+
+    def get_summary(self, raw_attributes):
+        """ returns the suitable instrument based on the 'abstract' attribute """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aperrin66: OK, it's good that you fixed the duplicates coming from pythesint, but you still need to fix this.
As it is, the test will always return True, so several normalizers can add the same value to the cumulative parameter.
You need to check that each element in the `new_member` list is not contained in the `output_cumulative_parameters[param]` list.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+        if set(['abstract']).issubset(raw_attributes.keys()):
+            return raw_attributes['abstract']
+        else:
+            return None
+
+    def get_provider(self, raw_attributes):
+        """Returns a GCMD-like provider from data structure"""
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Container class of get_ methods for osisaf normalizer """
+
+    def get_instrument(self, raw_attributes):
+        """ returns the suitable instrument based on the 'instrument_type' attribute (priortized one)
+        and 'activity_type' attribute """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            provider = None
+        return provider
+
+    def get_location_geometry(self, raw_attributes):
+        """Returns a GEOSGeometry object corresponding to the location of the dataset"""
+        if set(['northernsmost_latitude', 'southernmost_latitude',
+                'easternmost_longitude', 'westernmost_longitude']).issubset(raw_attributes.keys()):
+            polygon = utils.wkt_polygon_from_wgs84_limits(
+                # notice the difference between "northernSmost_latitude" and "northernmost_latitude" of default normalizer
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_netcdfcf.py
```diff
@@ -0,0 +1,20 @@
...
+class OSISAFMetadataNormalizer(unittest.TestCase):
+    """Tests for the NETCDFCF normalizer"""
+
+    def setUp(self):
+        self.normalizer = normalizers.NETCDFCFMetadataNormalizer([], [])
+
+    @mock.patch.object(pti, 'search_cf_standard_name_list',return_value=[{'standard_name':'sea_ice_area_fraction', 'amip':'sic'}])
+    def test_dataset_parameters(self, mock_normalize):
+        """ dataset_parameters from OSISAFMetadataNormalizer. Shall return the proper value
+         (one of "concentration", "type" or "drift") """
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_netcdfcf.py
```diff
@@ -0,0 +1,20 @@
...
+        attributes = {'raw_dataset_parameters': ['sea_ice_y_displacement','sea_ice_x_displacement']}
+
+        self.assertEqual(self.normalizer.get_dataset_parameters(attributes)[0],
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_netcdfcf.py
```diff
@@ -0,0 +1,20 @@
...
+import unittest.mock as mock
+import pythesint as pti
+
+class OSISAFMetadataNormalizer(unittest.TestCase):
+    """Tests for the NETCDFCF normalizer"""
+
+    def setUp(self):
+        self.normalizer = normalizers.NETCDFCFMetadataNormalizer([], [])
+
+    @mock.patch.object(pti, 'search_cf_standard_name_list',return_value=[{'standard_name':'sea_ice_area_fraction', 'amip':'sic'}])
```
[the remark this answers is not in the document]
aanersc: @aperrin66 
Because pythesint gives us a very long fields inside the returned value that is not favourable for testing purposes, I have mocked it. It is based on one of your previous comments. I have changed it to test values.
Done in the next commit.

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,191 @@
...
+from collections import OrderedDict
+
+import dateutil
+from dateutil.tz import tzutc
+from django.contrib.gis.geos.geometry import GEOSGeometry
+import metanorm.normalizers as normalizers
+
+
+class OSISAFMetadataNormalizer(unittest.TestCase):
+    """Tests for the ACDD attributes normalizer"""
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/test_handler.py
```diff
@@ -11,21 +11,45 @@
...
+            """Geospatial metadata handler"""
+            NORMALIZERS = [
+                normalizers.NETCDFCFMetadataNormalizer,
+                normalizers.SentinelSAFEMetadataNormalizer,
+                normalizers.SentinelOneIdentifierMetadataNormalizer,
+                normalizers.ACDDMetadataNormalizer,
+            ]
+        with self.assertRaises(ValueError):
+            testHandler(['test_output_parameter'], ['test_cumulative_parameter'])
+
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/handlers.py
```diff
@@ -20,13 +20,18 @@ class MetadataHandler():
...
+            raise ValueError(
+                "'BaseDefaultMetadataNormalizer' class must be inherited for the last one in the list of normalizers in 'MetadataHandler' class")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/test_handler.py
```diff
@@ -11,21 +11,45 @@
...
 
+    def test_GeoSpatialDefault_normalizer_be_last_in_chain(self):
+        """It is mandatory to have 'GeoSpatialDefaultMetadataNormalizer' in the list of normalizers (the last one).
+        This normalizer should never be removed from the list of normalizers(otherwise, ValueError must be raised)"""
+        handler = handlers.GeospatialMetadataHandler(
+            ['test_output_parameter'], ['test_cumulative_parameter'])
+        self.assertTrue(
+            normalizers.geospatial_defaults.GeoSpatialDefaultMetadataNormalizer == handler.NORMALIZERS[-1])
+
+        class testHandler(handlers.MetadataHandler):
```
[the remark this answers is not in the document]
aanersc: Sorry. Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

metanorm/normalizers/base.py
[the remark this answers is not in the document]
aanersc: Done in the next commit.

aanersc:
@aperrin66 
All of your comments are addressed in 5792235.

metanorm/normalizers/NETCDF_CF.py
```diff
@@ -0,0 +1,25 @@
...
+    raw_dataset_parameters attribute """
+
+    def get_dataset_parameters(self, raw_attributes):
+        """ Get the dataset's parameters """
+        if set(['raw_dataset_parameters']).issubset(raw_attributes.keys()):
+            standardized_dataset_parameters = list()
+            for raw_parameter_name in raw_attributes['raw_dataset_parameters']:
+                if pti.search_cf_standard_name_list(raw_parameter_name):
+                    standardized_dataset_parameters.append(
+                        pti.search_cf_standard_name_list(raw_parameter_name)[0])
```
aperrin66, line 24: Replace the use of the `search_cf_standard_name_list()` method by a utils function that checks wkv, then cf variables, and replace the `if` block by a `try` ... `except` block.

metanorm/normalizers/NETCDF_CF.py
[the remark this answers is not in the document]
aperrin66: Did you push it? The name is still the same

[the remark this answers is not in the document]
aanersc: I have pushed it. But it seems git does not track the changes in this regard!!!

See below:
image
from this link:
https://stackoverflow.com/questions/10523849/changing-capitalization-of-filenames-in-git


I think I have to force push next time!

aanersc:
@aperrin66 
Regarding changing the file name, I finally make it by online version git(github) in cd61bed. Today's discussion is addressed in d2bedcb. the file name is changed in cd61bed

metanorm/normalizers/base.py
```diff
@@ -25,35 +36,69 @@ def next(self):
...
+            try:
+                new_members = getattr(self, 'get_' + param)(raw_attributes)
+                for new_member in new_members:
+                    if new_member not in output_cumulative_parameters[param]:
+                        output_cumulative_parameters[param].append(new_member)
+            except AttributeError:
+                LOGGER.debug("%s: no method available for the '%s' parameter",
+                             self.__class__.__name__, param)
+            except TypeError:
+                LOGGER.debug("%s: method of the '%s' parameter does not working properly",
```
aperrin66, line 85: This slipped my notice before, but this message is too vague.
You need to explain when TypeErrors happen.

But should they even be caught here? If there is this kind of problem with a `get_` method, maybe it's best the the exception is not ignored.

metanorm/utils.py
```diff
@@ -147,3 +147,10 @@ def geometry_from_wkt_string(wkt_string, srid=4326):
     EPSG:4326, i.e. WGS 84
     """
     return GEOSGeometry(wkt_string, srid=srid)
+
+def two_list_search(keyword):
+    try:
+        result_list = pti.search_cf_standard_name_list(keyword)
```
aperrin66, line 153: you need to use `pti.get_cf_standard_name()`, then the result is not a list

metanorm/utils.py
```diff
@@ -147,3 +147,10 @@ def geometry_from_wkt_string(wkt_string, srid=4326):
...
+    except IndexError:
+        result_list = pti.get_wkv_variable(keyword)
```
aperrin66, line 155: `pti.get_wkv_variable()` does not return a list

metanorm/normalizers/base.py
aanersc: Do you mean removing this line?

I think it is better to have it. Since this type of harvesting parameters must return an array in their return statement (unlike the normal ones).

metanorm/utils.py
```diff
@@ -147,3 +147,10 @@ def geometry_from_wkt_string(wkt_string, srid=4326):
     EPSG:4326, i.e. WGS 84
     """
     return GEOSGeometry(wkt_string, srid=srid)
+
+def two_list_search(keyword):
```
aperrin66, line 151: That name is too vague. Something like `get_cf_or_wkv_standard_name()`?

aanersc: DOne in the next commit

tests/normalizers/test_netcdfcf.py
```diff
@@ -0,0 +1,22 @@
...
+import unittest.mock as mock
+import pythesint as pti
+
+class OSISAFMetadataNormalizer(unittest.TestCase):
+    """Tests for the NETCDFCF normalizer"""
+
+    def setUp(self):
+        self.normalizer = normalizers.NETCDFCFMetadataNormalizer([], [])
+
+    @mock.patch.object(pti, 'search_cf_standard_name_list',
```
aperrin66, line 13: you will need to update this since you now use the utils function in the netcdf_cf normalizer

metanorm/normalizers/netcdf_cf.py
```diff
@@ -0,0 +1,30 @@
...
+    raw_dataset_parameters attribute """
+
+    def get_dataset_parameters(self, raw_attributes):
+        """ Get the dataset's parameters """
+        if set(['raw_dataset_parameters']).issubset(raw_attributes.keys()):
+            standardized_dataset_parameters = list()
+            for raw_parameter_name in raw_attributes['raw_dataset_parameters']:
+                try:
+                    result = utils.two_list_search(raw_parameter_name)
+                except:
```
aperrin66, line 24: This is almost never a good idea: you will catch all exceptions, even unforeseen ones.
Please catch `IndexError`, which is the exception raise by pti methods when they don't find anything.

tests/test_handler.py
```diff
@@ -11,21 +11,49 @@
...
 
+    def test_GeoSpatialDefault_normalizer_be_last_in_chain(self):
+        """It is mandatory to have 'GeoSpatialDefaultMetadataNormalizer' in the list of normalizers
+        (the last one)."""
+        handler = handlers.GeospatialMetadataHandler(
+            ['test_output_parameter'], ['test_cumulative_parameter'])
+        self.assertTrue(
+            normalizers.geospatial_defaults.GeoSpatialDefaultMetadataNormalizer
+            == handler.NORMALIZERS[-1]
+            )
```
aperrin66, line 22: `self.assertEqual()`?

tests/normalizers/test_netcdfcf.py
aanersc: DOne in the next commit

metanorm/utils.py
aanersc: This suggestion harms the tests and the procedure of the code. I think it should remain like this, why do you suggest it? If we want to do it that way, then we have to have something like this for this function:
image

which tests are not passed for it:
ex1.
image

ex2.

image

metanorm/normalizers/base.py
aperrin66: If I understand well, a TypeError happens when the attribute method does not return an array.
In that case, an error should happen and be visible so that the behavior of this method can be fixed.
If you catch the error and only write a log message, the code will continue executing and the error will be much less visible, which strongly decreases the chance of the code being fixed.

metanorm/utils.py
aperrin66: You are mixing up two types of method from pythesint:
- `pti.search_...` methods return a list of matching parameter names, or an empty list if does not find anything
- `pti.get_...` methods return a single parameter name which perfectly matches the argument it is given, and raises an IndexError if it does not find it.

Here, we want to find a standard name that matches perfectly matches the input, so we should only use `get_...` methods.
So, something like this:
```python
def get_cf_or_wkv_standard_name(keyword):
    try:
        return pti.get_cf_standard_name(keyword)
    except IndexError:
        return pti.get_wkv_variable(keyword)
```

metanorm/normalizers/netcdf_cf.py
aanersc: Done in the next commit.

tests/test_handler.py
aanersc: Done in the next commit.

metanorm/normalizers/base.py
aanersc: I removed it. Done in the next commit.

aanersc: Once again after runing the tests for harvesting repo. I have found that why I put it there. This execpt is for the cases that new_members is None and causes this error. So, for that cases, this line should be here. So the code would be something like this:
```python
            try:
                new_members = getattr(self, 'get_' + param)(raw_attributes)
                for new_member in new_members:
                    if new_member not in output_cumulative_parameters[param]:
                        output_cumulative_parameters[param].append(new_member)
            except AttributeError:
                LOGGER.debug("%s: no method available for the '%s' parameter",
                             self.__class__.__name__, param)
            except TypeError: # for the cases that "new_members" is become None
                LOGGER.debug("%s: method of the '%s' parameter does not return any new values",
                             self.__class__.__name__, param)
```

aperrin66: yeah but `new_members` should not be None: the cumulative attribute methods should return an empty list if they find nothing

aanersc: Done in the next commit.

aanersc:
@aperrin66 
Regarding this comment of yours:https://github.com/nansencenter/metanorm/pull/14#discussion_r465089782
We have two append. One in the `netcdf-cf` normalizer and one in the `base.py`
These two ables us to return the element of pti in the `util.get_cf_or_wkv_standard_name` function for further processing. So there is no need to return a list at the `util.get_cf_or_wkv_standard_name` function level.

Everything is ready to merge in 7dfbaff if you agreed.

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,99 @@
...
+from .base import BaseMetadataNormalizer
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class OSISAFMetadataNormalizer(BaseMetadataNormalizer):
+    """ Normalizer for the attributes of datasets provided by OSISAF """
+
+    def get_instrument(self, raw_attributes):
```
aperrin66, line 20: Don't you need to adapt this according to @akorosov 's comment [here](https://github.com/nansencenter/metanorm/issues/10#issuecomment-668571305)?

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,99 @@
...
+        if set(['instrument_type']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                # make it unknown for osisaf products only
+                return utils.get_gcmd_instrument(utils.UNKNOWN)
+        else:
+            return None
+
+    def get_platform(self, raw_attributes):
```
aperrin66, line 34: Don't you need to adapt this according to @akorosov 's comment [here](https://github.com/nansencenter/metanorm/issues/10#issuecomment-668571305)?

aanersc: Done in 054c596

metanorm/normalizers/osisaf.py
aanersc: Done in 054c596

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,114 @@
...
+        Returns the suitable instrument based on the 'instrument_type' attribute (prioritized one)
+        and 'activity_type' attribute
+        """
+        if set(['instrument_type']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_instrument(raw_attributes['instrument_type'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                if ('_ice_conc' in raw_attributes['product_name']
+                or '_ice_type' in raw_attributes['product_name']
+                or '_ice_edge' in raw_attributes['product_name']):
```
aperrin66, line 32:
```suggestion
                    or '_ice_type' in raw_attributes['product_name']
                    or '_ice_edge' in raw_attributes['product_name']):
```

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,114 @@
...
+                    return pti.get_gcmd_instrument('AMSR2')
+                elif '_lr_ice_drift' in raw_attributes['product_name']:
+                    return pti.get_gcmd_instrument('Earth Remote Sensing Instruments')
+                elif '_mr_ice_drift' in raw_attributes['product_name']:
+                    return pti.get_gcmd_instrument('AVHRR')
+                ###elif  in raw_attributes['product_name']:
+                ###    return pti.get_gcmd_instrument('Imaging Spectrometers/Radiometers')
+                ###elif  in raw_attributes['product_name']:
+                ###    return pti.get_gcmd_instrument('Imaging Spectrometers/Radiometers')
+                # make it unknown for osisaf products only
```
aperrin66, line 44: please remove this commented code

metanorm/normalizers/osisaf.py
```diff
@@ -0,0 +1,114 @@
...
+            return None
+
+    def get_platform(self, raw_attributes):
+        """ returns the suitable platform based on the 'platform_name' attribute (prioritized one)
+        and 'activity_type' attribute """
+        if set(['platform_name']).issubset(raw_attributes.keys()):
+            return utils.get_gcmd_platform(raw_attributes['platform_name'])
+        elif set(['product_name']).issubset(raw_attributes.keys()):
+            if 'osi_saf' in raw_attributes['product_name']:
+                # make it unknown for osisaf products only
```
aperrin66, line 56: it's not unknown anymore

metanorm/normalizers/osisaf.py
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
aanersc: Done in the next commit

metanorm/normalizers/osisaf.py
aanersc: Done in the next commit

aanersc:
@aperrin66 
734a704 is ready if you agreed.

tests/normalizers/test_base.py
```diff
@@ -35,8 +37,9 @@ def get_other_parameter(self, raw_attributes):
 
     def test_instantiation(self):
         """Test the instantiation of a MetadataNormalizer"""
-        normalizer = self.TestMetadataNormalizer(['test_parameter'])
-        self.assertListEqual(normalizer._parameter_names, ['test_parameter'])
+        normalizer = self.TestMetadataNormalizer(['test_parameter'], [])
+        self.assertListEqual(
+            normalizer._output_parameters_names, ['test_parameter'])
```
aperrin66, line 43: You need to add a test to check that when both arguments of the constructor are None, a ValueError is raised.

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,256 @@
...
+            self.normalizer.get_instrument(attributes),
+            OrderedDict([('Category', 'Unknown'),
+                         ('Class', 'Unknown'),
+                         ('Type', 'Unknown'),
+                         ('Subtype', 'Unknown'),
+                         ('Short_Name', 'Unknown'),
+                         ('Long_Name', 'Unknown')])
+        )
+        attributes = {'test': 'test'}
+        self.assertIsNone(self.normalizer.get_instrument(attributes))
```
aperrin66, line 102: `test_instrument` can be split into multiple test methods

tests/normalizers/test_osisaf.py
```diff
@@ -0,0 +1,256 @@
...
+
+        attributes = {'product_name': 'osi_saf_2'}
+        # in the absence of 'platform_name' value should be UNKNOWN with GCMD template
+        self.assertEqual(
+            self.normalizer.get_platform(attributes),
+            OrderedDict([('Category', 'Earth Observation Satellites'),
+                         ('Series_Entity', ''),
+                         ('Short_Name', ''),
+                         ('Long_Name', '')])
+        )
```
aperrin66, line 123: `test_platform` can be split into multiple test methods

tests/normalizers/test_base.py
aanersc: Done in the next commit.

tests/normalizers/test_osisaf.py
aanersc: Done in the next commit.

tests/normalizers/test_osisaf.py
aanersc: Done in the next commit.

aanersc:
@aperrin66 
ee0e4ed is ready if you agreed.

aanersc:
@aperrin66 
It is ready I think 5d7ba3a is ready. It is 100 percent coverage and revised by autopep8!

metanorm/normalizers/base.py
```diff
@@ -25,35 +36,66 @@ def next(self):
...
                     LOGGER.debug("%s: value not found for the '%s' parameter",
                                  self.__class__.__name__, param)
 
+        for param in output_cumulative_parameters.keys():
+            previous_len = len(output_cumulative_parameters[param])
+            try:
+                new_members = getattr(self, 'get_' + param)(raw_attributes)
+                for new_member in new_members:
+                    if new_member not in output_cumulative_parameters[param]:
+                        output_cumulative_parameters[param].append(new_member)
```
akorosov, line 80: Which line is actually rising `AttributeError`?
Only this line should be after `try` other lines should be before `try` or after `else`

metanorm/normalizers/base.py
```diff
@@ -63,22 +105,23 @@ class BaseDefaultMetadataNormalizer(BaseMetadataNormalizer):
...
         """
-        Loops through parameters and checks if a method is available for that parameter.
-        If so use it,
+        Loops through output_parameters and checks if a method is available for that parameter.
+        If so use it, if not raise a MetadataNormalizationError
         """
-        if not parameters:
-            parameters = dict({(key, None) for key in self._parameter_names})
+        if not output_parameters:
+            output_parameters = dict({(key, None) for key in self._output_parameters_names})
```
akorosov, line 114: So we can define normalizer as
`n = BaseDefaultMetadataNormalizer(['some_param'])`
and then use it as
`n.normalize(attributes, ['another_params'])`
and in this case `some_param` will be ignored, right?
Why is this needed?

PR #25, opened 2020-09-09
Issue23 no django

### LINKED ISSUE #23: Metanorm depends on Django
Normalizers create instance of GeosGeometry and return to Harvesting, but that creates a dependency on Django. To reduce dependecies (and make Metanorm easier integrated in other software, e.g. Nansat) GeosGeometry should be created in Harvesters and Metanorm should return only strings.

akorosov:
Closes #23 
GEOSGeometry and create_geometry is removed. Now metanorm will return only WKT string.

### DIFF
--- a/metanorm/normalizers/acdd.py
+++ b/metanorm/normalizers/acdd.py
@@ -59,7 +59,7 @@ class ACDDMetadataNormalizer(BaseMetadataNormalizer):
             return None
 
     def get_location_geometry(self, raw_attributes):
-        """Returns a GEOSGeometry object corresponding to the location of the dataset"""
+        """Returns a WKT string corresponding to the location of the dataset"""
 
         if set(['geospatial_lat_max', 'geospatial_lat_min',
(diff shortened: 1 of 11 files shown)


aanersc (approved):
That is good to separate the things are not needed to be dependant on each other. But there are some tiny modifications of harvesting repo files must be done in order to makes things ready to work with text geometry. 
Here is the line in the test that must be changed:
https://github.com/nansencenter/django-geo-spaas-harvesting/blob/665aa4f102c591de6c9ae9b6959dba77d30c9e54/tests/test_ingesters.py#L430
And here is the line in the source cpde that must be changed:
https://github.com/nansencenter/django-geo-spaas-harvesting/blob/665aa4f102c591de6c9ae9b6959dba77d30c9e54/geospaas_harvesting/ingesters.py#L103

akorosov:
Yes, I have addressed that in this PR: and have already asked you to review it:
https://github.com/nansencenter/django-geo-spaas-harvesting/pull/29/files

PR #27, opened 2020-09-09
Issue26 radarsat2 csv

### LINKED ISSUE #26: Add normalizer for Radarsat2 CSV file from CSA
I've got a CSv file from Canada with metadata like that:
```
Result Number,Satellite,Date,Beam Mode,Polarization,Type,Image Id,Image Info,Metadata,Reason,Sensor Mode,Orbit Direction,Order Key,SIP Size (MB),Service UUID,Footprint,Look Orientation,Band,Title,Options,Absolute Orbit,Orderable
1,RADARSAT-2,2020-07-16 02:14:27 GMT,ScanSAR Wide A (W1 W2 W3 S7),HH HV,SGF,831163,"{""headers"":[""Product Type"" ""LUT Applied"" ""Sampled Pixel Spacing (Panchromatic)"" ""Product Format"" ""Geodetic Terrain Height""] ""relatedProducts"":[{""values"":[""SGF"" ""Ice"" ""100.0"" ""GeoTIFF"" ""0.00186""]}] ""collectionID"":""Radarsat2"" ""imageID"":""7337877""}",dummy value,,ScanSAR Wide,Ascending,RS2_OK121511_PK1076349_DK1021326_SCWA_20200716_021427_HH_HV_SGF,84,SERVICE-RSAT2_001-000000000000000000,-146.008396 73.905427 -143.459486 72.212173 -127.936480 73.451549 -128.875274 75.249738 -146.008396 73.905427 ,Right,C,rsat2_20200716_N7370W13656,,65706,TRUE
```

I want to make a small ingester for this kind of files (maybe part of harvesting) but I need metanorm to interpret this metadat correcttly.

akorosov:
Fixes #26 
Adds new normalizer and tests

### DIFF
--- a/metanorm/normalizers/radarsat2_csv.py
+++ b/metanorm/normalizers/radarsat2_csv.py
@@ -29,8 +29,10 @@ class Radarsat2CSVMetadataNormalizer(BaseMetadataNormalizer):
                 raw_attributes['Date'], '%Y-%m-%d %H:%M:%S GMT').replace(tzinfo=tzutc())
 
     def get_time_coverage_end(self, raw_attributes):
-        """ return end data from start date and add 5 minutes"""
-        return self.get_time_coverage_start(raw_attributes) + dt.timedelta(minutes=5)
+        """ return end data from start date and add 5 minutes """
+        time_coverage_start = self.get_time_coverage_start(raw_attributes)
+        if time_coverage_start:
+            return time_coverage_start + dt.timedelta(minutes=5)
 
     def get_platform(self, raw_attributes):
         """ return Radrasat-2 platform """
--- a/metanorm/normalizers/radarsat2_csv.py
+++ b/metanorm/normalizers/radarsat2_csv.py
@@ -74,7 +74,7 @@ class Radarsat2CSVMetadataNormalizer(BaseMetadataNormalizer):
     def get_dataset_parameters(self, raw_attributes):
         """ return list with sigma0 parameter from wkv variable """
         if self.match_metadata(raw_attributes):
-            return [utils.get_cf_or_wkv_standard_name('surface_backwards_scattering_coefficient_of_radar_wave')]
+            return [pti.get_wkv_variable('surface_backwards_scattering_coefficient_of_radar_wave')]
         else:
             return []
 
--- a/metanorm/normalizers/radarsat2_csv.py
+++ b/metanorm/normalizers/radarsat2_csv.py
@@ -1,4 +1,4 @@
-"""Normalizer for the interpretation of file name convention"""
+"""Normalizer metadat from CSV file from Radarsat2 facility"""
 
 import datetime as dt
 import logging
--- a/metanorm/normalizers/radarsat2_csv.py
+++ b/metanorm/normalizers/radarsat2_csv.py
@@ -1,0 +1,83 @@
+"""Normalizer for the interpretation of file name convention"""
+
+import datetime as dt
+import logging
+
+from dateutil.tz import tzutc
+import pythesint as pti
+
+from .base import BaseMetadataNormalizer
+import metanorm.utils as utils
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.addHandler(logging.NullHandler())
+
+
+class Radarsat2CSVMetadataNormalizer(BaseMetadataNormalizer):
+    """ Normalizer for metadata from CSV from Radarsat-2 searching facility """
+
+    def match_metadata(self, raw_attributes):
+        """ Check if metadata fits the normalizer """
+        keys = set(['Satellite', 'Date', 'Order Key', 'Footprint'])
+        return (keys.issubset(set(raw_attributes.keys()))
+            and raw_attributes['Satellite'] == 'RADARSAT-2')
+
+    def get_time_coverage_start(self, raw_attributes):
+        """ return start date from Date """
+        if self.match_metadata(raw_attributes):
+            return dt.datetime.strptime(
+                raw_attributes['Date'], '%Y-%m-%d %H:%M:%S GMT').replace(tzinfo=tzutc())
+
+    def get_time_coverage_end(self, raw_attributes):
+        """ return end data from start date and add 5 minutes"""
+        return self.get_time_coverage_start(raw_attributes) + dt.timedelta(minutes=5)
+
+    def get_platform(self, raw_attributes):
+        """ return Radrasat-2 platform """
+        if self.match_metadata(raw_attributes):
+            return pti.get_gcmd_platform('RADARSAT-2')
+
+    def get_instrument(self, raw_attributes):
+        """ return C-SAR instrument """
+        if self.match_metadata(raw_attributes):
+            return pti.get_gcmd_instrument('C-SAR')
+
+    def get_provider(self, raw_attributes):
+        """ return C-SAR instrument """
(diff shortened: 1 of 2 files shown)


tests/normalizers/test_radarsat2_csv.py
```diff
@@ -0,0 +1,136 @@
...
+        """ shall return RS2 image name """
+        self.assertEqual(
+            self.normalizer.get_entry_id(self.metadata),
+            'RS2_OK121511_PK1076349_DK1021326_SCWA_20200716_021427_HH_HV_SGF')
+
+    def test_get_parameters(self):
+        """ shall return sigma0 wkv """
+        self.assertEqual(
+            self.normalizer.get_dataset_parameters(self.metadata)[0]['standard_name'],
+            'surface_backwards_scattering_coefficient_of_radar_wave')
```
aanersc, line 114: Compare the whole ordered dictionary not only the standard of the first element. why the first element?

tests/normalizers/test_radarsat2_csv.py
```diff
@@ -0,0 +1,136 @@
...
+        'Service UUID,Footprint,Look Orientation,Band,Title,Options,Absolute Orbit,Orderable')
+        values = ('1,RADARSAT-2,2020-07-16 02:14:27 GMT,ScanSAR Wide A (W1 W2 W3 S7),HH HV,SGF,'+
+        '831163,"{""headers"":[""Product Type"" ""LUT Applied"" ""Sampled Pixel Spacing '+
+        '(Panchromatic)"" ""Product Format"" ""Geodetic Terrain Height""] ""relatedProducts"":'+
+        '[{""values"":[""SGF"" ""Ice"" ""100.0"" ""GeoTIFF"" ""0.00186""]}] ""collectionID"":"'+
+        '"Radarsat2"" ""imageID"":""7337877""}",dummy value,,ScanSAR Wide,Ascending,'+
+        'RS2_OK121511_PK1076349_DK1021326_SCWA_20200716_021427_HH_HV_SGF,84,'+
+        'SERVICE-RSAT2_001-000000000000000000,-146.008396 73.905427 -143.459486 72.212173 '+
+        '-127.936480 73.451549 -128.875274 75.249738 -146.008396 73.905427 ,Right,C,'+
+        'rsat2_20200716_N7370W13656,,65706,TRUE')
```
aanersc, line 25: Write them in a readable format. It is impossible to see which value belongs to which name here in this writing style.

metanorm/normalizers/radarsat2_csv.py
```diff
@@ -0,0 +1,85 @@
...
+        if self.match_metadata(raw_attributes):
+            return [utils.get_cf_or_wkv_standard_name('surface_backwards_scattering_coefficient_of_radar_wave')]
+        else:
+            return []
+
+    def get_entry_title(self, raw_attributes):
+        """ return list with sigma0 parameter from wkv variable """
+        if self.match_metadata(raw_attributes):
+            keys = ['Title', 'Polarization', 'Beam Mode']
+            return ' '.join([raw_attributes[key] for key in keys])
```
aanersc, line 85: you can simple return a string if it is not changing! no need to use `join`

metanorm/normalizers/radarsat2_csv.py
```diff
@@ -0,0 +1,85 @@
...
+        """ return C-SAR instrument """
+        if self.match_metadata(raw_attributes):
+            return pti.get_gcmd_provider('CSA')
+
+    def get_location_geometry(self, raw_attributes):
+        """ return a WKT string corresponding to the location of the dataset"""
+        if self.match_metadata(raw_attributes):
+            fp = raw_attributes['Footprint'].strip().split(' ')
+            return (f'POLYGON(({fp[0]} {fp[1]}, {fp[2]} {fp[3]}, {fp[4]} {fp[5]}, ' +
+            f'{fp[6]} {fp[7]}, {fp[8]} {fp[9]}))')
```
aanersc, line 57: If you are 100 percent sure that 'Footprint' has nine element then it is correct. If not, then add more code to check that in order not to receive an error.

aanersc (commented):
FInd out the comments please.

metanorm/normalizers/radarsat2_csv.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_entry_id(self, raw_attributes):
+        """ return filename """
+        if self.match_metadata(raw_attributes):
+            return raw_attributes['Order Key']
+
+    def get_dataset_parameters(self, raw_attributes):
+        """ return list with sigma0 parameter from wkv variable """
+        if self.match_metadata(raw_attributes):
+            return [utils.get_cf_or_wkv_standard_name('surface_backwards_scattering_coefficient_of_radar_wave')]
```
aanersc, line 77: although `get_cf_or_wkv_standard_name` function facilitates the value finding between the two list, I strongly suggest you to find out the value from the correct list and call that method of that list. This speeds up normalization process. This way of coding slows down every cases that we normalizer. However, the demand frequency of demand of radarsat cases are not know to me. It up to you.`get_cf_or_wkv_standard_name` function has two try/expect statements that are an overhead calculation which is not necessary at all.

tests/normalizers/test_radarsat2_csv.py
```diff
@@ -109,9 +124,15 @@ def test_get_entry_id(self):
...
-            'surface_backwards_scattering_coefficient_of_radar_wave')
+        self.assertCountEqual(
+            self.normalizer.get_dataset_parameters(self.metadata),
+            [OrderedDict([('standard_name',
+              'surface_backwards_scattering_coefficient_of_radar_wave'),
+             ('long_name', 'Normalized Radar Cross Section'),
+             ('short_name', 'sigma0'),
+             ('units', 'm/m'),
+             ('minmax', '0 0.1'),
+             ('colormap', 'gray')])])
```
aanersc, line 135: using `OrderedDict` and `assertCountEqual` at the same time is contradictory!! Please change the `assertCountEqual`. We use `OrderedDict` from `collections` in order to preserve the order of those elemets! It should not be tested with an assertion that ignores the ordering.

akorosov:
@aanersc , yes, I saw it. assertCountEqual is applied to __lists__

tests/normalizers/test_radarsat2_csv.py
akorosov: assertCountEqual compares two __lists__. One is output from `get_dataset_parameters ` another __list__ contains `OrderedDict `. I think it should be fine.

metanorm/normalizers/radarsat2_csv.py
akorosov: It always ten elements

tests/normalizers/test_radarsat2_csv.py
akorosov: Done

tests/normalizers/test_radarsat2_csv.py
akorosov: Done

tests/normalizers/test_radarsat2_csv.py
aanersc: fine. Yeah you are right.The `get_dataset_parameters` must return a list unlike other get_... methods.

aanersc (approved):

PR #28, opened 2020-09-10
Issue22 entry id config

aanersc:


aanersc:
I don't know why but irrelevant commit are also added to this pr. Please, pay attention to the last commit only!!!!

metanorm/normalizers/url.py
```diff
@@ -192,5 +196,12 @@ def get_location_geometry(self, raw_attributes):
...
 
     def get_entry_title(self, raw_attributes):
-        """ returns the suitable provider based on the filename """
+        """ returns the suitable entry_title based on the filename """
         return self.find_matching_value(self.urls_title, raw_attributes)
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        if 'url' in raw_attributes:
+            if any(raw_attributes['url'].startswith(key) for key in self.urls_title.keys()):
```
akorosov, line 205: What if URL come from DDX ingester? Then this will not work. Please prepare a more generic solution and discuss with me before implementing.

akorosov (changes requested):
URL-normalizer should also be able to process any url from DDX ingester. With the current implementation it is not possible. Please prepare a solution that works with DDX and discuss with me before further development.

aanersc:
@akorosov 
YOur comment is addressed in ce4da89.

akorosov (changes requested):
Great! Can you also add examples of files from podaac and osisaf to tests?

akorosov (approved):

akorosov:
@aperrin66 , can you have a look? Now metanorm should return entry_id from raw_attributes from all ingesters.

metanorm/normalizers/url.py
```diff
@@ -185,12 +193,19 @@ def create_parameter_list(parameters):
...
     def get_entry_title(self, raw_attributes):
-        """ returns the suitable provider based on the filename """
+        """ returns the suitable entry_title based on the filename """
         return self.find_matching_value(self.urls_title, raw_attributes)
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        if 'url' in raw_attributes:
+            if any(raw_attributes['url'].startswith(url_start) for url_start in self.urls_entry_id):
+                file_name = os.path.splitext(os.path.basename(raw_attributes['url']))[0]
```
aperrin66, line 210: `os.path.splitext()` only splits the last extension.
So for a link like those we get from thredds.met.no: 

https://thredds.met.no/thredds/dodsC/osisaf/met.no/ice/drift_mr/2014/07/ice_drift_nh_polstere-200_avhrr-ch2_201407112204_201407122158.nc.dods

We will get the following result: `ice_drift_nh_polstere-200_avhrr-ch2_201407112204_201407122158.nc`

But for a link from PODAAC like this:

https://opendap.jpl.nasa.gov/opendap/allData/ghrsst/data/GDS2/L2P/MODIS_A/JPL/v2019.0/2020/085/20200325000000-JPL-L2P_GHRSST-SSTskin-MODIS_A-D-v02.0-fv01.0.nc

We will get `20200325000000-JPL-L2P_GHRSST-SSTskin-MODIS_A-D-v02.0-fv01.0`.

aperrin66, line 210: Although it works, be aware that using `os.path.basename()` on a URL is not what this function was intended for.
You are taking advantage of an undocumented behavior, so it's not guaranteed that it will always work in later versions.
Also, I don't know how that would behave on Windows.

tests/normalizers/test_url.py
```diff
@@ -305,6 +305,41 @@ def test_entry_title_ceda(self):
...
+        attributes = {
+            'url': 'ftp://anon-ftp.ceda.ac.uk/neodc/esacci/sst/data/CDR_v2/Climatology/L4/v2.1/D365-ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_CDR2.1-v02.0-fv01.0.nc'}
+        self.assertEqual(self.normalizer.get_entry_id(
+            attributes), 'D365-ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_CDR2.1-v02.0-fv01.0')
+
+
+    def test_entry_id_for_osisaf_ingester(self):
+        """entry_id from URLMetadataNormalizer for osisaf project """
+        attributes = {
+            'url': "https://thredds.met.no/thredds/catalog/osisaf/Some/path/to/file/filename.bb"}
```
aperrin66, line 332: the download URL would be "https://thredds.met.no/thredds/catalog/osisaf/Some/path/to/file/filename.bb.dods"

aperrin66 (changes requested):
There are a few things to change and a few elements to be aware of, otherwise it looks fine.

metanorm/normalizers/url.py
```diff
@@ -185,12 +193,19 @@ def create_parameter_list(parameters):
...
+                return file_name if file_name else None
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/normalizers/test_sentinel_safe.py
```diff
@@ -287,3 +287,20 @@ def test_provider_is_none_for_non_scihub_url(self):
     def test_provider_missing_attribute(self):
         """Parameter method must return None if the attribute is missing"""
         self.assertEqual(self.normalizer.get_provider({}), None)
+
+    def test_entry_id_copernicus(self):
+        """entry_id from sentinelSafeMetadataNormalizer """
+        attributes = {
+            'url': "https://scihub.copernicus.eu/apihub/odata/v1/Products('1a4ff15b-1504-4d94-8675-e12c06b02858')/$value",
+            'Identifier':'finename_value'}
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

metanorm/normalizers/url.py
aanersc: @aperrin66 
SO, how many times should we split? one time for podaac and two times for osisaf?
shall we use another if here?

aanersc: @aperrin66 
splitext is the suggestion of @akorosov . I would rather use the split method of python. what is your opinion @akorosov ?

tests/normalizers/test_url.py
aanersc: Done in the next commit.

metanorm/normalizers/url.py
aperrin66: This might not even be a problem, I am just raising the point to make sure it has been considered.

aanersc: @aperrin66 
I have tested that this line is also working:
```python
file_name = os.path.basename(raw_attributes['url']).split('.')[0]
```
which one should I put here? above one or the splitext one?

aanersc: Done in the next commit.

aanersc:
@aperrin66 
@akorosov 
Everything except the exclusivity comment is solved in the 1cb465b. If you agreed,it is ready to merge.otherwise....

metanorm/normalizers/url.py
```diff
@@ -185,12 +193,23 @@ def create_parameter_list(parameters):
...
+        """ returns the suitable entry_title based on the filename """
         return self.find_matching_value(self.urls_title, raw_attributes)
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        file_name = None
+        if 'url' in raw_attributes:
+            if any(raw_attributes['url'].startswith(url_start) for url_start in self.urls_entry_id):
+                file_name = os.path.splitext(os.path.basename(raw_attributes['url']))[0] or \
+                    ntpath.split(ntpath.basename(raw_attributes['url']))[0] # for windows users
```
aperrin66, line 212: That won't work for several reasons:
- `os.path.basename()` will return something on Windows, I just wasn't sure how it would treat the `/` characters in the URL since the filesystem separator in Windows is `\`. So your second line is never executed.
- the second line does not do what you want anyway: `ntpath` is the Windows implementation of `os.path`, so it uses `\`, so it won't work on URLs

aanersc:
@aperrin66
@akorosov
Everything except the exclusivity comment is solved in the be2bb1d. If you agreed,it is ready to merge.otherwise....

metanorm/normalizers/url.py
```diff
@@ -185,12 +193,24 @@ def create_parameter_list(parameters):
...
+                extension_set={'.nc','.h5'}
+                for ext in extension_set: # removing the extensions
+                    file_name = file_name.replace(ext, '')
```
aperrin66, line 215: That's too specific, please keep one of the two solutions you mentioned before.
`split('.')` is the most generic one here, so I think I'd go with that if @akorosov agrees.

aperrin66 (changes requested):

metanorm/normalizers/url.py
aanersc: @aperrin66 
`splitext` is actually a one-time `split('.')`!!! nothing more! So, @akorosov shall decide. If we have cases like `blablabla.nc.dods` there is no other than that code to remove the `nc` in the second round of extension removal. This is all related to how frequently we encounter such situation.

aanersc: @aperrin66 
@akorosov 
something that should be mentioned is that when we use `split('.')` we can't see the rest of the filename if there is a version number specified with '.'. For example, 'f35_20140603v8.2' filename is always being read as 'f35_20140603v8'

aperrin66: > splitext is actually a one-time split('.')!!! nothing more! 

Not exactly:
- `'blablabla.nc.dods'.split('.')` returns `['blablabla', 'nc', 'dods']`
- `os.path.splitext('blablabla.nc.dods')` returns `('blablabla.nc', '.dods')`

So `file_name.split('.')[0]` will remove all extensions whatever the number, whereas `os.path.splitext(file_name)` only removes the last extension.

aperrin66: In order to have something that works every time, `urls_entry_id` can be turned into a dictionary in which the URL is the key and the value is a regular expression that can be used to extract the file name (or any part of the URL that we want to use as ID).

aanersc:
@aanersc 
Your comment is addressed in 793690c.

metanorm/normalizers/url.py
```diff
@@ -48,6 +50,12 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
         "ftp://ftp.remss.com/gmi/": 'Atmosphere parameters from Global Precipitation Measurement Microwave Imager',
         "ftp://ftp.gportal.jaxa.jp/standard": 'AMSR2-L2 Sea Surface Temperature'}
 
+    urls_entry_id = {"https://thredds.met.no/thredds/catalog/osisaf/met.no/ice":lambda x: re.findall(r"^ice_.{39}",x),
+                     "https://opendap.jpl.nasa.gov/opendap/":lambda x: re.findall(r"\d{14}.{62}",x),
+                     "ftp://ftp.remss.com/":lambda x: re.findall(r"f35_.{16}|f35_.{12}|f35_.{10}",x),
+                     "ftp://ftp.gportal.jaxa.jp":lambda x: re.findall(r".{6}_.{34}",x),
+                     "ftp://anon-ftp.ceda.ac.uk/":lambda x: re.findall(r"D\d{3}-.{56}",x)
+                     }
```
aperrin66, line 58: No need to use lambdas, just store [compiled regular expressions](https://docs.python.org/3/library/re.html#re.compile) here.

aperrin66, line 58: Just use `search()` instead of `findall()[0]`

aanersc: Done in the next commit

aanersc: Done in the next commit

aanersc:
@aperrin66 
YOur comment is addressed in e7db6e8.

I have also carefully modified the other dictionary of entry_title for the exact assignment of title.

aanersc:
There is a minor modification in 637f3a5 based on pylint.

metanorm/normalizers/url.py
```diff
@@ -46,8 +47,20 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
...
         "ftp://ftp.remss.com/gmi/": 'Atmosphere parameters from Global Precipitation Measurement Microwave Imager',
-        "ftp://ftp.gportal.jaxa.jp/standard": 'AMSR2-L2 Sea Surface Temperature'}
-
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L2.SST/": 'AMSR2-L2 Sea Surface Temperature',
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L3.SST_10/": 'AMSR2-L2 Sea Surface Temperature',
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L3.SST_25/": 'AMSR2-L2 Sea Surface Temperature'
+        }
+
+    urls_entry_id = {"https://thredds.met.no/thredds/catalog/osisaf/met.no/ice":
+                     re.compile(r"^ice_.{39}"),
```
aperrin66, line 56: Using the number of characters is very rigid and will stop working at the slightest change.
You also have the opportunity here to get rid of the problematic `os.path.basename()` call below.

For this particular case, you could use something like `r'([^/]+)\.nc\.dods$'`.

Then, you can apply it to the whole URL in `get_entry_id()` (you will need to use `self.urls_entry_id[url_start].search(...).group(1)` to extract the file name).

metanorm/normalizers/url.py
```diff
@@ -185,12 +198,22 @@ def create_parameter_list(parameters):
...
         return self.find_matching_value(self.urls_title, raw_attributes)
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        file_name = None
+        if 'url' in raw_attributes:
+            for url_start in self.urls_entry_id:
+                if raw_attributes['url'].startswith(url_start):
+                    file_name = re.search(self.urls_entry_id[url_start],
+                                          os.path.basename(raw_attributes['url'])).group(0)
```
aperrin66, line 218: the regex search needs to be inside a `try` block

aanersc:
@aperrin66 
coded as you guided in ea1803d.

aanersc:
@aperrin66 
I have also added the second possible extension that could be in that specific ftp resource.

aanersc:
@aperrin66 
it seems one more test is needed! based on coverage!

aanersc:
@aperrin66 
the test is added in 8c45444.

metanorm/normalizers/url.py
```diff
@@ -46,7 +47,20 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
...
         "ftp://ftp.remss.com/gmi/": 'Atmosphere parameters from Global Precipitation Measurement Microwave Imager',
-        "ftp://ftp.gportal.jaxa.jp/standard": 'AMSR2-L2 Sea Surface Temperature'}
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L2.SST/": 'AMSR2-L2 Sea Surface Temperature',
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L3.SST_10/": 'AMSR2-L2 Sea Surface Temperature',
+        "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/L3.SST_25/": 'AMSR2-L2 Sea Surface Temperature'
+        }
+
+    urls_entry_id = {"https://thredds.met.no/thredds/catalog/osisaf/met.no/ice":
+                     re.compile(r"([^/]+)\.nc\.dods"),
+                     "https://opendap.jpl.nasa.gov/opendap/": re.compile(r"([^/]+)\.nc|([^/]+)\.gz"),
```
aperrin66, line 57: No need to repeat the whole pattern: `r"([^/]+)\.(nc|gz)"` does the same thing.
Plus, if there is a `.gz` extension, it's probably not going to be by itself, more likely `.nc.gz`, so this is probably more relevant: `r"([^/]+)\.nc.(gz)?"`
Please fix all the regexes accordingly.

metanorm/normalizers/url.py
```diff
@@ -46,7 +47,20 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
...
+                     "ftp://ftp.remss.com/gmi": re.compile(r"([^/]+)\.gz|([^/]+)\.nc"),
+                     "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/":
+                     re.compile(r"([^/]+)\.h5|([^/]+)\.nc"),
+                     "ftp://anon-ftp.ceda.ac.uk/neodc/esacci/sst/data/CDR_v2/Climatology/":
+                     re.compile(r"([^/]+)\.nc|([^/]+)\.h5")
```
aperrin66, line 62: Please add a `$` anchor at the end of all these regular expressions.

metanorm/normalizers/url.py
```diff
@@ -185,12 +199,26 @@ def create_parameter_list(parameters):
...
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        file_name = None
+        if 'url' in raw_attributes:
+            for url_start in self.urls_entry_id:
+                if raw_attributes['url'].startswith(url_start):
+                    try:
+                        file_name = \
+                        re.search(self.urls_entry_id[url_start],raw_attributes['url']).group(1) \
+                        or re.search(self.urls_entry_id[url_start],raw_attributes['url']).group(2)
```
aperrin66, line 221: you can remove the second line once you applied the suggestion of the previous comment

aanersc:
@aperrin66 
addressed in 40cffc3.

metanorm/normalizers/url.py
```diff
@@ -185,12 +201,25 @@ def create_parameter_list(parameters):
...
+
+    def get_entry_id(self, raw_attributes):
+        """ returns the suitable entry_id based on the filename """
+        file_name = None
+        if 'url' in raw_attributes:
+            for url_start in self.urls_entry_id:
+                if raw_attributes['url'].startswith(url_start):
+                    try:
+                        file_name = re.search(
+                            self.urls_entry_id[url_start],raw_attributes['url']).group(1)
```
aperrin66, line 222: The standard way of doing this would be: `self.urls_entry_id[url_start].search(raw_attributes['url']).group(1)`

aperrin66 (approved):

PR #36, opened 2020-10-28
More generic entry_id regex for thredds.met.no

### LINKED ISSUE #35: Make the thredds.met.no entry_id regex more generic
Related to https://github.com/nansencenter/django-geo-spaas-harvesting/issues/44

The regex used to extract the entry_id from thredds.met.no URLs should work for any URL starting with "https://thredds.met.no/thredds/".

aperrin66:
Resolves #35 
Related to nansencenter/django-geo-spaas-harvesting#44

Also cleaned up the dictionaries indentation in the url normalizer, which might make the diff hard to read.
The only significant change in url.py is on line 85.

### DIFF
--- a/metanorm/normalizers/url.py
+++ b/metanorm/normalizers/url.py
@@ -82,7 +82,7 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
     }
 
     urls_entry_id = {
-        "https://thredds.met.no/thredds/dodsC/osisaf/met.no/ice": re.compile(r"([^/]+)\.nc\.dods$"),
+        "https://thredds.met.no/thredds/": re.compile(r"([^/]+)\.nc(\.dods)?$"),
         "https://opendap.jpl.nasa.gov/opendap/": re.compile(r"([^/]+)\.(nc|h5)(\.gz)?$"),
         "ftp://ftp.remss.com/gmi": re.compile(r"([^/]+)\.gz$"),
         "ftp://ftp.gportal.jaxa.jp/standard/GCOM-W/GCOM-W.AMSR2/":
--- a/metanorm/normalizers/url.py
+++ b/metanorm/normalizers/url.py
@@ -18,13 +18,16 @@ LOGGER.addHandler(logging.NullHandler())
 
 class URLMetadataNormalizer(BaseMetadataNormalizer):
     """ Normalizer for hardcoding information based on URLS """
+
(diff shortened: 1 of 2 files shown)


aanersc:
Caution!!
Don't we want another product from `https://thredds.met.no/thredds/` rather than osisaf? If something else is needed from `https://thredds.met.no/thredds/` then this code makes everything entry_id related to the ice and ice concentration, etc.

If we don't want anything else except osisaf, then this code is ok, otherwise it is incorrect set the values for other non-relevant product as well!

@akorosov 
@aperrin66

aperrin66:
> Caution!!
> Don't we want another product from `https://thredds.met.no/thredds/` rather than osisaf? If something else is needed from `https://thredds.met.no/thredds/` then this code makes everything entry_id related to the ice and ice concentration, etc.
> 
> If we don't want anything else except osisaf, then this code is ok, otherwise it is incorrect set the values for other non-relevant product as well!
> 
> @akorosov
> @aperrin66

The new regex extracts the file name from the URL for any product, I don't see the problem.

For example, if it gets this URL:
https://thredds.met.no/thredds/fileServer/met.no/observations/stations/SN99938.nc

Then `URLMetadataNormalizer.get_entry_id()` will return "SN99938".

(Also please use the "Review changes" functionality with the "Comment" option for this kind of remark, it makes it easier to answer).

aanersc:
> > Caution!!
> > Don't we want another product from `https://thredds.met.no/thredds/` rather than osisaf? If something else is needed from `https://thredds.met.no/thredds/` then this code makes everything entry_id related to the ice and ice concentration, etc.
> > If we don't want anything else except osisaf, then this code is ok, otherwise it is incorrect set the values for other non-relevant product as well!
> > @akorosov
> > @aperrin66
> 
> The new regex extracts the file name from the URL for any product, I don't see the problem.
> 
> For example, if it gets this URL:
> https://thredds.met.no/thredds/fileServer/met.no/observations/stations/SN99938.nc
> 
> Then `URLMetadataNormalizer.get_entry_id()` will return "SN99938".
> 
> (Also please use the "Review changes" functionality with the "Comment" option for this kind of remark, it makes it easier to answer).

Yes, now I remember, it only sets the entry_id, nothing more!

aanersc (approved):

akorosov (approved):
Very good!

PR #42, opened 2020-11-27
Issue40 url time coverage

### LINKED ISSUE #40: Refactor the URL normalizer's time coverage code
The code dealing with time coverage in the URL normalizer is unreadable and repetitive.
It should be rewritten in a clearer way.

There are also mistakes to fix in the time coverage for some sources.

aperrin66:
Resolves #40 

The information about time coverage for each URL is now stored in a single location: the `urls_time` dictionary.

Also fixed some wrong time coverage ends (see unit tests).

### DIFF
--- a/metanorm/normalizers/url.py
+++ b/metanorm/normalizers/url.py
@@ -290,8 +290,7 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
                 for matcher, get_time, get_coverage in time_finder:
                     match = matcher.search(raw_attributes['url'])
                     if match:
-                        time_info = match.groupdict()
-                        file_time = get_time(**time_info)
+                        file_time = get_time(**match.groupdict())
                         return (get_coverage(file_time)[0], get_coverage(file_time)[1])
         return (None, None)
 
--- a/metanorm/normalizers/url.py
+++ b/metanorm/normalizers/url.py
@@ -10,6 +10,7 @@ import pythesint as pti
 from dateutil.relativedelta import relativedelta
 from dateutil.tz import tzutc
 
+import metanorm.utils as utils
 from .base import BaseMetadataNormalizer
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


metanorm/normalizers/url.py
```diff
@@ -205,124 +249,59 @@ def get_instrument(self, raw_attributes):
...
-                # the constant date is corrected using the day
-                # number in the beginning of the file name
-                extracted_date += relativedelta(days=int(file_name[1:4]) - 1)
+            time_finder = self.find_matching_value(self.urls_time, raw_attributes)
+            if time_finder:
+                for matcher, get_time, get_coverage in time_finder:
+                    match = matcher.search(raw_attributes['url'])
+                    if match:
+                        time_info = match.groupdict()
+                        file_time = get_time(**time_info)
```
aanersc, line 294: These two lines can be merged into:
```suggestion
                        file_time = get_time(**match.groupdict())
```

aanersc (commented):
For the sake of `ease to use` and `more meaningful metadata time designing`, it is better to subtract the infinitesimal amount of time from  the period of dataset at the end of coverage of the dataset! It means if a user points to a specific point of time as the starting point of her/his desired time period, s/he not see the results that ends exactly at that point of time. This requires the second elemet of third function which you added as a lambda function becomes `+ relativedelta(hours=23, minutes=59, seconds=59)` instead of `+ relativedelta(days=1)`. Because of that one second, the result will not appear if the point of time that comes from the user is exactly the end of period of time of data. This is more meaningful than seeing it in the result. This also should apply for `+ relativedelta(months=1)` and the other intervals for `time_coverage_end`. It is incorrect to do it for `time_coverage_start` because exact starting point of time should be in the coverage to bring that dataset in the search results. I am happy to discuss it further.

tests/normalizers/test_url.py
```diff
@@ -136,15 +136,15 @@ def test_time_coverage_start_multiobs_glo_phy_nrt_015_003_hourly(self):
         """shall return the propert starting time for hardcoded normalizer """
         self.assertEqual(
             self.normalizer.get_time_coverage_start(
-                {'url': 'ftp://nrt.cmems-du.eu/Core/MULTIOBS_GLO_PHY_NRT_015_003/dataset-uv-nrt-hourly/2020/09/dataset-uv-nrt-hourly_20200906T1800Z_P20200918T0000.nc'}),
-            datetime(year=2020, month=9, day=6, hour=18, minute=0, second=0, tzinfo=tzutc()))
+                {'url': 'ftp://nrt.cmems-du.eu/Core/MULTIOBS_GLO_PHY_NRT_015_003/dataset-uv-nrt-hourly/2020/09/dataset-uv-nrt-hourly_20200906T0000Z_P20200912T0000.nc'}),
+            datetime(year=2020, month=9, day=6, hour=0, minute=0, second=0, tzinfo=tzutc()))
```
aanersc, line 140: It is better to have that value 20200906T1800Z in the test in order to see that 18 comes out of result of normalizer. 00 is not a good choice here.

aperrin66:
@aanersc regarding https://github.com/nansencenter/metanorm/pull/42#pullrequestreview-540834824

We need to store the exact time coverage. If we do what you suggest, we are missing one second.
It is the responsibility of the code using the data to make the comparisons properly. To avoid the case you describe, simply using strict inequalities is enough.

aanersc:
> @aanersc regarding [#42 (review)](https://github.com/nansencenter/metanorm/pull/42#pullrequestreview-540834824)
> 
> We need to store the exact time coverage. If we do what you suggest, we are missing one second.
> It is the responsibility of the code using the data to make the comparisons properly. To avoid the case you describe, simply using strict inequalities is enough.

@aperrin66 
no human will blame us for loosing one second, but we are sure that the correct anticipation of users are satisfied. This code store the the values in the database, so it is better to prevenet cousing the problems from the fountain! :)

If we include that ending point, then we always have to be careful of using `<=`or `<`! everywhere! if we correct here, then we can code in the future without any concerns. This is my suggestion, it is up to you to decide.

tests/normalizers/test_url.py
aperrin66: For these datasets, "hourly" qualifies the resolution, not the time coverage.
The time in these datasets' names is always 0000, because they contain 6-hourly data covering a full day, from midnight to midnight. So even if there was another value, it should not be taken into account.

metanorm/normalizers/url.py
aperrin66: Done

metanorm/utils.py
```diff
@@ -157,3 +159,27 @@ def get_cf_or_wkv_standard_name(keyword):
...
+    """
+    year = int(year)
+    hour = int(hour)
+    minute = int(minute)
+    second = int(second)
+
+    if day_of_year:
+        day_of_year = int(day_of_year)
+        first_day = datetime(year, 1, 1, hour, minute, second).replace(tzinfo=tzutc())
+        return first_day + timedelta(days=day_of_year-1)
```
aanersc, line 181: consider spaces for -1

aanersc (approved):

akorosov (approved):
Great!

PR #45, opened 2020-12-02
Add information in the summary field

### LINKED ISSUE #43: Add information to the summary field
The summary field should include:
- [ ] the processing level (L1, L2, etc...)
- [ ] for models, the name of the model

These should be added with a structure eliminating possible ambiguities when appending to an existing summary.

aperrin66:
Resolves #43.
I limited the modifications to the products necessary for TOPVOYS.

### DIFF
--- a/metanorm/normalizers/acdd.py
+++ b/metanorm/normalizers/acdd.py
@@ -25,8 +25,16 @@ class ACDDMetadataNormalizer(BaseMetadataNormalizer):
 
     def get_summary(self, raw_attributes):
         """Get the dataset's summary"""
-        if set(['summary']).issubset(raw_attributes.keys()):
-            return raw_attributes['summary']
+        summary_fields = []
+
+        if 'summary' in raw_attributes.keys():
+            summary_fields.append(f"Description: {raw_attributes['summary']}")
+
+            if 'processing_level' in raw_attributes.keys():
+                processing_level = raw_attributes['processing_level'].lstrip('L')
+                summary_fields.append(f"Processing level: {processing_level}")
+
+            return ';'.join(summary_fields)
         else:
             return None
 
--- a/metanorm/normalizers/acdd.py
+++ b/metanorm/normalizers/acdd.py
@@ -25,16 +25,16 @@ class ACDDMetadataNormalizer(BaseMetadataNormalizer):
 
     def get_summary(self, raw_attributes):
         """Get the dataset's summary"""
-        summary_fields = []
+        summary_fields = {}
 
         if 'summary' in raw_attributes.keys():
-            summary_fields.append(f"Description: {raw_attributes['summary']}")
+            summary_fields[utils.SUMMARY_FIELDS['description']] = raw_attributes['summary']
 
             if 'processing_level' in raw_attributes.keys():
                 processing_level = raw_attributes['processing_level'].lstrip('Ll')
-                summary_fields.append(f"Processing level: {processing_level}")
+                summary_fields[utils.SUMMARY_FIELDS['processing_level']] = processing_level
 
-            return ';'.join(summary_fields)
+            return utils.dict_to_string(summary_fields)
         else:
             return None
 
--- a/metanorm/normalizers/acdd.py
+++ b/metanorm/normalizers/acdd.py
@@ -31,7 +31,7 @@ class ACDDMetadataNormalizer(BaseMetadataNormalizer):
             summary_fields.append(f"Description: {raw_attributes['summary']}")
 
             if 'processing_level' in raw_attributes.keys():
-                processing_level = raw_attributes['processing_level'].lstrip('L')
+                processing_level = raw_attributes['processing_level'].lstrip('Ll')
                 summary_fields.append(f"Processing level: {processing_level}")
 
             return ';'.join(summary_fields)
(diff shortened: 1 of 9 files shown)


metanorm/normalizers/acdd.py
```diff
@@ -25,8 +25,16 @@ def get_entry_title(self, raw_attributes):
...
         """Get the dataset's summary"""
-        if set(['summary']).issubset(raw_attributes.keys()):
-            return raw_attributes['summary']
+        summary_fields = []
+
+        if 'summary' in raw_attributes.keys():
+            summary_fields.append(f"Description: {raw_attributes['summary']}")
+
+            if 'processing_level' in raw_attributes.keys():
+                processing_level = raw_attributes['processing_level'].lstrip('L')
```
aanersc, line 34:
```suggestion
                processing_level = raw_attributes['processing_level'].lstrip('L').lstrip('l')
```

aperrin66: Why not, but there is no need to call `lstrip()` twice. `lstrip('Ll')` does the job.

tests/normalizers/test_sentinel_safe.py
```diff
@@ -23,19 +23,64 @@ def test_entry_title_missing_attribute(self):
         """Parameter method must return None if the attribute is missing"""
         self.assertEqual(self.normalizer.get_entry_title({}), None)
 
-    def test_summary(self):
+    def test_summary_description_only(self):
         """summary from SentinelSAFEMetadataNormalizer"""
         attributes = {
             'Date': '2018-04-18T01:02:03Z',
-            'Instrument': 'instrument_value',
+            'Instrument name': 'instrument_value',
```
aanersc, line 30: Is this a bug that you fixed? I mean, is 'Instrument name' correct and 'Instrument' never seen in the raw attributes?

metanorm/normalizers/acdd.py
aperrin66: done

tests/normalizers/test_sentinel_safe.py
aperrin66: No, it's just that we already use the "Instrument" field in `get_instrument()`, so it's more useful to have the full instrument name in the summary-

metanorm/normalizers/acdd.py
```diff
@@ -25,8 +25,16 @@ def get_entry_title(self, raw_attributes):
 
     def get_summary(self, raw_attributes):
         """Get the dataset's summary"""
-        if set(['summary']).issubset(raw_attributes.keys()):
-            return raw_attributes['summary']
+        summary_fields = []
+
+        if 'summary' in raw_attributes.keys():
+            summary_fields.append(f"Description: {raw_attributes['summary']}")
```
aanersc, line 31: it is better not to use `append` method of list because of speed! Especially in the metanorm. Do you have another way of coding for it which does not use `append`?

tests/normalizers/test_acdd.py
```diff
@@ -23,10 +23,21 @@ def test_entry_title_missing_attribute(self):
...
+    def test_summary_description_and_processing_level(self):
+        """summary from ACDDMetadataNormalizer, with the description and processing level"""
+        attributes = {
+            'summary': 'summary_value',
+            'processing_level': 'L2'
+        }
+        self.assertEqual(
+            self.normalizer.get_summary(attributes),
+            'Description: summary_value;Processing level: 2'
+        )
```
aanersc, line 40: So now, we always have `Processing level` part of the summary for all raw metadata that comes with acdd format(regardless of belonging to toopvoys or not)! Is it a correct behaviour?

metanorm/normalizers/acdd.py
aperrin66: It's not a problem to use `append()`.
Inserting and removing elements from the beginning of the list is slow, because all elements have to be shifted, but that's not what we do here.  https://docs.python.org/3.7/tutorial/datastructures.html#using-lists-as-queues

Plus, the list is very small, so even if we did that it would not be a major problem.

tests/normalizers/test_acdd.py
aperrin66: Yes

metanorm/normalizers/acdd.py
aanersc: fair enough.

tests/normalizers/test_acdd.py
aanersc: But you only want to change the topvoys ones! not the other ones!

aperrin66: No I don't. When I said in the PR description that I only modified the products necessary for TOPVOYS I meant that I did not take the time to do it for the other products we can harvest. It's not a problem to have extra information, even for products which are not used in TOPVOYS.

aanersc: So, if you want to formalise the summery, then it is better to do it in the `util` section of code and each normalizer fill-in their own parts. And also, we should do it in a way that not restricted to topvoys only!

aperrin66: I wasn't considering systematically giving a structure to the summary, but it would be more coherent. I'll give that some thought.

aperrin66: I made the summary generation a bit more generic by putting the summary fields names and the code which generates the string in utils.

aanersc (approved):

akorosov (approved):
Great!

PR #56, opened 2021-01-12
Update provider for remss.com

### LINKED ISSUE #55: NASA/GSFC/SED/ESD/LA/GPM seems to have been removed from GCMD providers
Replace it with REMSS

aperrin66:
Resolves #55

### DIFF
--- a/metanorm/normalizers/url.py
+++ b/metanorm/normalizers/url.py
@@ -43,7 +43,7 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
 
     urls_provider = {
         'ftp://anon-ftp.ceda.ac.uk/neodc/esacci/sst/data/CDR_v2/Climatology/L4/v2.1': 'ESA/CCI',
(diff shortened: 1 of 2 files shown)


metanorm/normalizers/url.py
```diff
@@ -43,7 +43,7 @@ class URLMetadataNormalizer(BaseMetadataNormalizer):
 
     urls_provider = {
         'ftp://anon-ftp.ceda.ac.uk/neodc/esacci/sst/data/CDR_v2/Climatology/L4/v2.1': 'ESA/CCI',
-        "ftp://ftp.remss.com/gmi/": 'NASA/GSFC/SED/ESD/LA/GPM',
```
aanersc, line 46: why do you remove gmi from the path of ftp? 
please be specific in metanorm about addressing. Otherwise in the future development it will cause semantic error (the code will run perfectly without any error and deliver the worng answers)which is extremely difficult to debug.

aperrin66: Because the "Remote sensing system" provider is valid for the whole REMSS FTP server, not just the gmi directory.

akorosov (approved):

PR #58, opened 2021-01-12
Update the test image

### LINKED ISSUE #57: The django package is no longer useful in the test image
Since #25, the `django` package is no longer required in the test image.

aperrin66:
Resolves #57 
Also switched to the slim nansat_base image for tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
     -e "TRAVIS_BRANCH=$TRAVIS_BRANCH"
     -e "TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST"
     "${TESTING_IMAGE_NAME}"
-    bash -c "source /opt/conda/bin/activate && coverage run --source=./metanorm -m unittest discover tests && coveralls"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,7 +32,7 @@ script:
     -e "TRAVIS_BRANCH=$TRAVIS_BRANCH"
     -e "TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST"
     "${TESTING_IMAGE_NAME}"
-    bash -c "source /opt/conda/bin/activate && coverage run --source=./metanorm -m unittest discover tests && coveralls"
```
aanersc, line 35: what is the rule of this part which you delete it?

aanersc: I mean `source /opt/conda/bin/activate`

aperrin66: The slim image is not based on conda, so there is no need to activate a conda environment anymore.

aanersc (approved):

akorosov (approved):

PR #61, opened 2021-02-04
Add a normalizer for CMEMS in situ TAC datasets

### LINKED ISSUE #54: Add support for CMEMS's INSITU_GLO_NRT_OBSERVATIONS_013_030
Add a normalizer for CMEMS's INSITU_GLO_NRT_OBSERVATIONS_013_030 product

aperrin66:
Resolves #54 

I did not find a way to makes this a generic normalizer and had to use the "id matching" technique.

### DIFF
--- a/tests/normalizers/test_cmems_in_situ_tac.py
+++ b/tests/normalizers/test_cmems_in_situ_tac.py
@@ -19,9 +19,9 @@ class CMEMSInSituTACMetadataNormalizerTestCase(unittest.TestCase):
         """matches_identifier() should return True if the
         id of the dataset is a CMEMS TAC id.
         """
-        self.assertTrue(self.normalizer.matches_identifier({'id': 'AB_CD_EF_foo_20201120'}))
-        self.assertTrue(self.normalizer.matches_identifier({'id': 'AB_CD_EF_Foo-Bar_20201120'}))
-        self.assertTrue(self.normalizer.matches_identifier({'id': 'AB_CD_EF_3EBE3_201812'}))
+        self.assertTrue(self.normalizer.matches_identifier({'id': 'NO_TS_TG_OsloTG_20210124'}))
+        self.assertTrue(self.normalizer.matches_identifier({'id': 'GL_TS_DB_5100628'}))
+        self.assertTrue(self.normalizer.matches_identifier({'id': 'AR_TS_MO_Blakksnes_201812'}))
 
         self.assertFalse(self.normalizer.matches_identifier({'id': 'A_B_C_D_E'}))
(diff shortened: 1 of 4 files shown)


aanersc (dismissed):

tests/normalizers/test_cmems_in_situ_tac.py
```diff
@@ -0,0 +1,205 @@
...
+        self.assertTrue(self.normalizer.matches_identifier({'id': 'AB_CD_EF_Foo-Bar_20201120'}))
+        self.assertTrue(self.normalizer.matches_identifier({'id': 'AB_CD_EF_3EBE3_201812'}))
+
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'A_B_C_D_E'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'AB_CD_EF'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': '1B_CD_EF_foo_20201120'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'AB_C2_EF_foo_20201120'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'AB_CD_34_foo_20201120'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'AB_CD_EF_foo_20201120_'}))
+        self.assertFalse(self.normalizer.matches_identifier({'id': 'AB_CD_EF_foo_20201120_bar'}))
```
aanersc, line 32: @aperrin66 Although I have accepted, one aspect comes into my mind afterwards. The exclusivity of normalizer should also be considered for this PR. Here I think the `'^[A-Z]{2}_[A-Z]{2}_[A-Z]{2}(_[^_]+){1,2}$'` regex pattern is not specific enough. The `'^[A-Z]{2}_[A-Z]{2}_[A-Z]{2}(_[^_]+){1,2}$'` is generic and might cause a problem in the future for catching the flow in filling the values that are not belong to this CMEMS TAC cases. Please use another regex, with more specific pattern that acts more precisely and exclusively and also modify the tests that become more realistic with more real values instead of AB, CD, EF,etc...

aperrin66: Sure, it does not hurt to be extra careful.
Done

aanersc: That is the semantic of code! Semantic of execution! It must be executed in the most accurate cases, no one else. This does not concerns being `extra careful`. Now we have 10 normalizers, think about the time that we have 25 normalizers and one of them incorrectly set the value for other inputs that does not belong to it! it would be very difficult to find it which one does this. So, the ONLY approach in writing our code in metanorm is to be EXTREMELY precise!

akorosov (approved):

PR #70, opened 2021-05-14
Updates to adapt to changes in pythesint

### LINKED ISSUE #69: The GCMD keywords for the Sentinel platforms changed again
They went from "capitalized" words ("Sentinel-1") to fully uppercase words ("SENTINEL-1").
We need to make sure that the output from `metanorm` is always the same, regardless of what `pythesint` provides.

aperrin66:
:warning: **Please read before merging.**

Resolves #69 and more

The GCMD vocabularies versions are pinned, so there is no need to worry about that for now, but we will need to update these in the future.

However, there has been changes to the structure of the data for the `cf_standard_name` vocabulary (see nansencenter/py-thesaurus-interface#36).
This means that pinning the version was not enough to ensure backwards compatibility.
I updated the tests to use the new values and data structure, but we need to think about the consequences in the existing databases before moving forward.

In general, we need to think about how to manage the evolution of the vocabularies.

For now this won't break anything, but the existing vocabularies using the old format will remain in existing databases.


akorosov:
I don"t know how to deal with changing of vocabularies. But I guess we need that fix already now. So I will approve. Adrien, can you suggest a meeting to discuss our approach? And maybe the approach?

akorosov (approved):

PR #82, opened 2021-07-30
Add support for Earthdata CMR search API

### LINKED ISSUE #80: Add support for Earthdata CMR metadata
Needed for nansencenter/django-geo-spaas-harvesting#58

Metadata specification: https://cdn.earthdata.nasa.gov/umm/granule/v1.6.3

aperrin66:
Resolves #80

### DIFF
--- a/metanorm/handlers.py
+++ b/metanorm/handlers.py
@@ -56,6 +56,7 @@ class GeospatialMetadataHandler(MetadataHandler):
         normalizers.SentinelSAFEMetadataNormalizer,
         normalizers.SentinelOneIdentifierMetadataNormalizer,
         normalizers.CreodiasEOFinderMetadataNormalizer,
+        normalizers.EarthdataCMRMetadataNormalizer,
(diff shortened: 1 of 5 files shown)


akorosov (approved):
Looks great! I only noticed that you use try/except instead of "if" to check several attributes. But in these cases it is probably much easier with try that testing nested dictionaries with 3, 4 layers. Right?

aperrin66:
> Looks great! I only noticed that you use try/except instead of "if" to check several attributes. But in these cases it is probably much easier with try that testing nested dictionaries with 3, 4 layers. Right?

I do that after testing that the metadata has the "umm" or "meta" key at the root level. If that is the case, then the metadata should have the expected structure and I use a try block.
I tried to explain that in [this comment](https://github.com/nansencenter/metanorm/blob/968dc98a29188320d5ad5ae3257d9f4694c7d8b1/metanorm/normalizers/earthdata_cmr.py#L24) but it might not be very clear.

